PR #14, opened 2017-10-29
python script improvement

nmizukami:
python preprocess script to remap gridded runoff to catchment hru average runoff

1. Implemented matrix multiplication to compute weighted average hru runoff from weight and gridded runoff time series array, instead of looping 

2. Implemented multiprocessing to parallelize hru average runoff computation processes.

3. use xarry for netcdf output.

4. a number of cleanups

### DIFF
--- a/scripts/Calc_hru_wgtavg_nc.py
+++ b/scripts/Calc_hru_wgtavg_nc.py
@@ -18,14 +18,17 @@ import netCDF4 as nc4
 #         hardcoded variables              #
 ############################################
 # number of processors (= number of chunks of hru to be processed)
-NPROC=4
-# Name of netCDF variable for polygon ID in input data
-HDIM='hru_id'
-TDIM='Time'
-# Name of netCDF variable for weight in weight data
+NPROC=2
+# Name of variables in runoff netCDF
+GRID_ID_NAME='hru_id'
+TIME_VAR_NAME='time'
+# Name of variables in weight netCDF
+HRU_ID_NAME='hruid'
 WGTNM='weight'
 OVRPLYNM='overlapPolyId'
 OVRNM='overlaps'
+
+FILL_VALUE=-9999.0
 ############################################ 
 #              Class                       #
 ############################################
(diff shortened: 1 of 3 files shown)


scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -89,115 +80,62 @@ def getNetCDFAtt(fn, varname,attName):
...
-          a = dataVal[10,ij[0][0],ij[1][0]].tolist()
-          #if value of overlapping polygon is missing data -case1
-          if a == None: 
-            numvoid = numvoid+1
-            wgtArray[j] = 0 
-    
-    #Initialize Val2[ntime] 
-    Val2 = np.zeros((dim1size))
+    a = np.zeros( (dim1size,len(wgtArray)) )
+    for j in range(overlaps):                     # Go through each overlapping polygon
```
jhamman: you can probably get rid of the `overlaps` variable too. 

```Python
for j, overlap_id in enumerate(overlap_Ids):
    ij = np.column_stack(np.where(ghruid == overlap_id))
```

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -89,115 +80,62 @@ def getNetCDFAtt(fn, varname,attName):
...
+      # find index of grid cell that match up with hru id of overlapsId[j]
+      ij=np.column_stack(np.where(ghruid==overlapsId[j]))
+      # if nc_in netCDF does not cover nc_wgt domain - case2
+      if len(ij[0]) == 0:
+        numvoid+=1
+        wgtArray[j]=0.0
+      else:
+        a[:,j] = dataVal[:,ij[0,0],ij[0,1]]
```
jhamman: This is some ugly indexing. What is going on here?

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -89,115 +80,62 @@ def getNetCDFAtt(fn, varname,attName):
...
-
-    #if there is no valid weight values (i.e. all zero)- nothing overlapping for current hru, so missing value
-    else:
-      Val2 = np.ones((dim1size))*-9999.0
-    # store weighted average time series, Val2 in final array wgtVal
-    wgtVal[:,i]= Val2
-    
-    del Val2
+      wgtArray = wgtArray/np.nansum(wgtArray)
+      wgtArray = np.where(np.isnan(wgtArray),0,wgtArray)
```
jhamman: what is the typical shape/rank of wgtArray? It may be faster to use bottlenecks `nansum` here. Also you compute the same value twice `np.nansum(wgtArray)` so save it as a variable.

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -89,115 +80,62 @@ def getNetCDFAtt(fn, varname,attName):
...
+      wgtVal[:,i]=np.dot(a,wgtArray.T).T
 
   return wgtVal
 
+def split(hrulist,n):
+  subsize=len(hrulist)/n
+  idx1=0
+  hrudata=[]
+  for i in range(n-1):
```
jhamman: since you don't use i here, use a dummy variable to make that clear:

```Python
for _ in range(n-1):
```

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -234,8 +172,21 @@ def usage():
...
-      wgtVal=compAvgVal(nc_wgt,nc_in,varname)
-      writeNetCDFData(nc_out, wgtVal, varname)     
+      func=partial(compAvgVal, nc_wgt, nc_in, varname)
+
+      pool = mp.Pool(processes=NPROC,maxtasksperchild=1)
+      results=pool.map(func,range(NPROC))
+      result=np.column_stack(results)
+
+      with xr.open_dataset(nc_in) as ds:
+        timedata  = ds.coords[TDIM]     # Get time coordinate variable
```
jhamman: you can probably drop the `coords` part here. You'll get the time variable either way but dropping `coords` will protect against a malformed input file that isn't decoded having the time variable as a coordinate.

jhamman: also, `TDIM`can probably be renamed to `TIME_VAR` or something more descriptive/accurate.

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -234,8 +172,21 @@ def usage():
...
+      with xr.open_dataset(nc_wgt) as ds:
+        hrudata = ds[HDIM].values
```
jhamman: why do you pull out the values here? Can you leave this as a DataArray?

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -234,8 +172,21 @@ def usage():
...
+
+      encoding={varname: {'dtype': 'float32', '_FillValue': -9999.0}}
```
jhamman: Use a module level variable for the fill value

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -234,8 +172,21 @@ def usage():
...
+      foo = xr.DataArray(result, coords=[('time',timedata), ('hru',hrudata)], name=varname)
+      foo.encoding = encoding[varname]
+      foo.to_netcdf('test.nc')
```
jhamman: just a reminder to provide a way to programmatically determine the output file name.

nmizukami: replaced 'test.nc' with nc_out

scripts/Calc_hru_wgtavg_nc.py
nmizukami: done

scripts/Calc_hru_wgtavg_nc.py
nmizukami: wgtArray is a vector with varying size for different hrus.  `np.nansum(wgtArray)` is saved as `sum_wgt`

scripts/Calc_hru_wgtavg_nc.py
nmizukami: Reason why I used `overlaps' (number of overlapping grid boxes) is overlap_Ids include nan  looking like `[ 32518.  32519.     nan     nan     nan     nan     nan     nan     nan]` and wanted to avoid looping through all the elements.    in this case, `overlaps` is 2

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -89,115 +80,62 @@ def getNetCDFAtt(fn, varname,attName):
...
 
   dataVal  = getNetCDFData(nc_in,varname) # Get data value 
   FillVal  = getNetCDFAtt(nc_in,varname,'_FillValue') # Get data value 
-  IdVal    = getNetCDFData(nc_in,IDNM) 
+  ghruid   = getNetCDFData(nc_in,HDIM)
   dim1size = dataVal.shape[0]
-
   #Initialize wgtVal[ntime,nhru] 
-  wgtVal = np.zeros((dim1size,len(hruIDs)))
+  wgtVal = np.ones((dim1size,len(hrulist)))*-9999.0
```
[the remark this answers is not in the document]
nmizukami: great suggestion!

scripts/Calc_hru_wgtavg_nc.py
jhamman: Ok, so why don't you just make `overlapsId` a variable that includes only valid elements.

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -89,115 +80,62 @@ def getNetCDFAtt(fn, varname,attName):
...
-  ncfile.source = os.path.dirname(os.path.abspath(__file__))+__file__[1:]
-  
-  ncfile.close()
-
-def compAvgVal(nc_wgt,nc_in,varname):
+def compAvgVal(nc_wgt, nc_in, varname, chunk):
   """Compute areal weighted avg value of <varname> in <nc_in> for each hru based on hru's weight in <nc_wgt>""" 
-  wgt = wgtnc(nc_wgt)             #instantaneous of wgtnc object
-  hruIdName = wgt.getHruIdName()  # Get precise name of variable for hru id 
+  wgt       = wgtnc(nc_wgt)       #instantaneous of wgtnc object
```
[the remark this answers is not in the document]
nmizukami: my bad english..

scripts/Calc_hru_wgtavg_nc.py
nmizukami: done

scripts/Calc_hru_wgtavg_nc.py
nmizukami: yes, ugly.   `dataVal` is [time, y, x] 3D runoff grid, and wanted to get indices for x and y. 
`ghruid`is [y,x] array including overlap_ids and get the indices from `ij=np.column_stack(np.where(ghruid==overlap_ids[j]))`   is there any better way (maybe rewrite in xarray)?

scripts/Calc_hru_wgtavg_nc.py
nmizukami: Ok,   this is what I do.   Remove nan element from 'overlap_ids' and 'wgt' array with `overlap_ids=overlap_ids[~np.isnan(overlap_ids)]` and the same for `wgt array`.   so I don't need `overlaps` variable and loop directly through `overlap_ids` array.

nmizukami:
I committed the changes based on specific suggestions.  As a separate pull request, I will replace wgtnc class to function.  and I will change style based on PEP8 (that is good guidance, since I don't have my particular style)

jhamman:
> and I will change style based on PEP8 (that is good guidance, and useful for me because I don't have my particular style)

take a look at flake8 and/or autopep8. You can probably get 90% of the way there having an automated linter do most of the work.

nmizukami:
@jhamman   I ran flake8 and clean up a little bit.   would you like to take a quick look? or just close this?

jhamman:
@nmizukami - looking much better. I think you could go further on the style improvements (in particular the CamelCase is not part of the pep8 standard) but that can wait for another day. Do you want to merge this in its current state?

nmizukami:
@jhamman yes, I would like to merge now (can you do it?   I don't think I have permission to to do).   Thanks for suggestions and comments!

jhamman:
@nmizukami See #16 and #17 for followup issues.

PR #15, opened 2017-10-30
Feature/restart files

nmizukami:
Split restart file into input and output


scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -86,11 +87,11 @@ def writeNetCDFData(fn, var, varname):
...
   dim2size=var.shape[1]
  
-  dim_1 = ncfile.createDimension('dim1',dim1size )  # hru axis
-  dim_2 = ncfile.createDimension('dim2',dim2size )  # hru axis
+  dim_1 = ncfile.createDimension(TDIM,dim1size )  # time axis
+  dim_2 = ncfile.createDimension(HDIM,dim2size )  # hru axis
 
   # Define a 2D variable to hold the var
-  val = ncfile.createVariable(varname,'f4',('dim1','dim2'))
+  val = ncfile.createVariable(varname,'f4',(TDIM,HDIM),zlib=True,complevel=1)
```
andywood: Would there be a reason to make compression level a user supplied parameter? Are the state files easily readable if not compressed, but not if they are compressed? Will uncompressed state files work better more easily with DA? Just some questions about usability later.

route/build/src/write_simoutput.f90
```diff
@@ -213,7 +213,8 @@ subroutine defineStateFile(fname,           &  ! input: filename
  ! define time bound 
  call defvar('time_bound', 'time bound at last time step',                'sec', (/sTB_DimName/),                               nf90_double, ierr, cmessage)
  ! Hill-Slope 
- call defvar('QFUTURE',    'Hill slope convoluted flow',                  'm3/s',(/sSeg_DimName,sTdh_DimName,sEns_dimName/),    nf90_double, ierr, cmessage)
+ call defvar('QFUTURE',    'Hill slope routed future flow',    'm3/s',(/sSeg_DimName,sTdh_DimName,sEns_dimName/),    nf90_double, ierr, cmessage)
```
andywood: changes look fine but it would be helpful to add a comment to say what 'future flow' and 'delayed runoff' are in the context of the routing.

route/build/src/kwt_route.f90
```diff
@@ -478,6 +482,18 @@ subroutine QROUTE_RCH(IENS,JRCH,    & ! input: array indices
     else
       ! set flow in headwater reaches to modelled streamflow from time delay histogram
       RCHFLX(IENS,JRCH)%REACH_Q = RCHFLX(IENS,JRCH)%BASIN_QR(1) 
+      if (allocated(KROUTE(IENS,JRCH)%KWAVE)) THEN
```
andywood: if there is a way to deallocate naturally at the end of a use of a variable, rather than have to check before allocating, I think that would be preferable.  in some cases that may be difficult to track, though.  also, if the reallocated variable has the same size as in its former use then just reusing it without de/re allocation will be more efficient.  I don't know if that's the case here.

scripts/Calc_hru_wgtavg_nc.py
nmizukami: Sorry Andy, I don't know why this python script change was sneaked into pull request.  This python script has been changed and merged in different pull request.     so

PR #19, opened 2017-11-10
Feature/restart writing mod

nmizukami:
1.  Extract state information stored in data types and store them in multi-dimension arrays.

2. Write state information from array without segment loops.


route/build/src/route_runoff.f90
```diff
@@ -121,6 +125,15 @@ program route_runoff
 type(KREACH),allocatable   :: wavestate(:,:)      ! wave states for all upstream segments at one time step
 integer(i4b),allocatable   :: wavesize(:,:)       ! number of wave for segments 
 integer(i4b)               :: LAKEFLAG            ! >0 if processing lakes
+! restart state variables
```
andywood: could add comment to define each of these variables (eg rf_array()) and their dims

andywood:
changes appear to make sense - using the 2D/3D writing fcns is efficient

PR #20, opened 2017-11-10
feature/restart_write_generic ncwrite

nmizukami:
1. created one module to store netcdf writing routine for different variable types and ranks.
2. use this generic netcdf writing routine in main program - runoff_route.f90


andywood:
separating this into a module makes sense
could be that moving parallel read_netcdf fcns into the same module would facilitate updating them together if anything has to change in the future

PR #22, opened 2017-11-13
Feature/python

nmizukami:
1. Variable name changes

2. Removed class (wgtnc) and implement functions instead


nmizukami:
@jhamman   I am not sure why conflicts show up in this PR.    I just continued modifying the python code on the top of the previous PR that was merged into develop branch.

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -1,208 +1,137 @@
...
+        (wgt_array, overlap_ids) = get_wgt(nc_wgt, hru_list[i])
+        # Go through wgt list and replace value with zero for following cases
+        # where overlapping polygon has missing value - case1
+        # where overlapping polygon is outside nc_wgt domain - case2
+        a = np.zeros((dim1size, len(wgt_array)))
+        for j, overlap_id in enumerate(overlap_ids):
+            # find index of grid cell that match up with hru id of overlap_ids
+            row, col = np.where(ghruid == overlap_id)
+            # if nc_in netCDF does not cover nc_wgt domain - case2
+            if not (np.size(row) and np.size(col)):
```
jhamman: are you just checking to see if these are empty arrays?

`if not row or not col` would be the cleaner way to write this.

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -1,208 +1,137 @@
...
-    del Val2
-
-  return wgtVal
-
-############################################
-#                Main                      #
-############################################
+                a[:, j] = np.squeeze(var_data[:, row, col])
+                # if value of overlapping polygon is missing data -case1
+                if any(a[:, j] == fill_val):
```
jhamman: this can just be `if fill_val in a[:, j]`

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -1,208 +1,137 @@
...
+                    wgt_array[j] = 0.0
+        sum_wgt = np.sum(wgt_array)
```
jhamman: nit: use the sum method on `wgt_array` when possible. Just a bit more "Pythonic"

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -1,208 +1,137 @@
...
+        if sum_wgt > 0.0:
+            # Adjust weight value if valid weight value (> 0) exist in list
+            wgt_array = wgt_array/sum_wgt
+            wgt_array = np.where(np.isnan(wgt_array), 0, wgt_array)
```
jhamman: I prefer to use `pd.isnull` in these cases. Its most cases, its more performant.

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -1,208 +1,137 @@
...
+
+def split(hru_list, n):
+    subsize = len(hru_list)/n
+    idx1 = 0
+    hru_data = []
+    for _ in range(n-1):
+        idx2 = idx1+subsize
+        hru_data.append(hru_list[idx1:idx2])
+        idx1 = idx2
+    hru_data.append(hru_list[idx1:])
```
jhamman: This should be refactored to use the range iterator correctly. I think that would look like this but you should test it.

```Python
+def split(hru_list, n):
    subsize = len(hru_list)/n
    hru_data = []
    for idx1 in range(0, n, subsize):
        idx2 = idx1 + subsize
        hru_data.append(hru_list[idx1: idx2])
    hru_data.append(hru_list[idx1:])
```

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -1,208 +1,137 @@
...
+
+# ----
+# Main
+# ----
 use = '''
-Usage: %s -[h] <weight_netCDF> <input_netCDF> <variable_name_in_input_netCDF> <output_netCDF>
+Usage: %s -[h] <weight_netCDF>
+               <input_netCDF>
+               <variable_name_in_input_netCDF>
+               <output_netCDF>
```
jhamman: just an idea, the `click` library makes for super easier command line interfaces. Might be worth checking out. 

http://click.pocoo.org/5/

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -216,9 +145,7 @@ def usage():
         usage()
 
     verbose = False
-    grid_info = False
-    proj_info = True
-    for (opt,val) in opts:
+    for (opt, val) in opts:
```
jhamman: parentheses are not necessary.

jhamman:
@nmizukami - I don't know why you have merge conflicts. Perhaps you branched off an old branch?

They don't look hard to correct. When you're making the changes I requested, just merge in master then and you'll have the opportunity to resolve the merge conflict then.

scripts/Calc_hru_wgtavg_nc.py
nmizukami: >are you just checking to see if these are empty arrays? 

That is correct, but did not get the same answer.

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -1,208 +1,137 @@
...
+
+def get_netCDF_data(fn, varname):
     """Read <varname> variables from NetCDF <fn> """
-    f = nc4.Dataset(fn,'r')
-    data = f.variables[varname][:]
+    f = nc4.Dataset(fn, 'r')
+    var_data = f.variables[varname][:]
     f.close()
-    return data
+    return var_data
```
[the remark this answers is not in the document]
nmizukami: I am testing this statement against previous netCDF read. For some reason, this xarray version is noticeably slower than using netcdf4 read (I don't know why) and also slightly different answer where some missing values exist in array.   I like this short statement, but I wanted to resolve these.

scripts/Calc_hru_wgtavg_nc.py
```diff
@@ -1,208 +1,137 @@
...
-          if (IdVal[yj,xi] != overlapsId[j]):
-             print "hru_id %d not exist, Sayonara" %IdVal[yj,xi]
-             sys.exit()
-          # Concatenate time series array, Val1 to oriVal containin time serise from all the intersected polygon 
-          if Val1.size: #if Val1 is not empty i.e., exclude empty value in weight list
-            c += 1
-            if c == 1:
-              oriVal = Val1
+    att_data = getattr(var, att_name)
+    return att_data
```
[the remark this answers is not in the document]
nmizukami: I am getting error when I read fill_values line 82.

scripts/Calc_hru_wgtavg_nc.py
nmizukami: Yes, I look into this.

scripts/Calc_hru_wgtavg_nc.py
nmizukami: I tested this and some modified versions, but I don't get the same answer (or getting error on index). I need to test this portion more separately.

nmizukami: Maybe something like this, but still does not quite produce what I want. You could check too.

```
def split1(some_list, nsplit):
    subsize = int(len(some_list)/nsplit)
    some_data = []
    for idx1 in range(0, len(some_list)-2*subsize, subsize):
        idx2 = idx1 + subsize
        some_data.append(some_list[idx1: idx2])
    some_data.append(some_list[idx2:])
    return some_data
```

maybe for now leave it is.

nmizukami:
Most of requests were incorporated, but i have questions on several of comments.

scripts/Calc_hru_wgtavg_nc.py
jhamman: Consider these examples:

```Python
In [13]: a = np.array([])

In [14]: b = np.array([])

In [15]: if not a or not b:
    ...:     print('neither a or b has any elements')
    ...:
neither a or b has any elements

In [16]: if not (a and b):
    ...:     print('neither a or b has any elements')
    ...:
neither a or b has any elements

In [17]: if not (a.size and b.size):
    ...:     print('neither a or b has any elements')
    ...:
neither a or b has any elements

In [18]: if not (np.size(a) and np.size(b)):
    ...:     print('neither a or b has any elements')
    ...:
neither a or b has any elements
```

If this doesn't give you what you want, I think I must be misunderstanding what you are trying to do.

scripts/Calc_hru_wgtavg_nc.py
jhamman: Can you provide an example of what the hru_list looks like? Now that I look at this again, you may be able to get away with something much simpler:

```Python
def split1(l, n):
    """Yield successive n-sized chunks from l."""
    for i in range(0, len(l), n):
        yield l[i:i + n]
```

ref: https://stackoverflow.com/a/312464/1757464

nmizukami:
```python
n_split = 4
n_list = 16
some_list = list(range(0,n_list))
print(some_list)
```
```
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]
```
```python
split_list = split(some_list, n_split)
print(split_list)
```
should be like 
```
[[0, 1, 2, 3], [4, 5, 6, 7], [8, 9, 10, 11], [12, 13, 14, 15]]
```

i will try stackoverflow example

scripts/Calc_hru_wgtavg_nc.py
nmizukami: @jhamman  I did a quick test.   Difference appears if you have element zero (which is valid index for python).     I think i will use `if not (a.size and b.size):`

```python

In [2]: import numpy as np

In [3]: a = np.array([0])

In [4]: b = np.array([1])

In [5]: print(a); print(b)
[0]
[1]

In [6]: if not a or not b:
   ...:     print('neither a or b has any elements')
   ...:     
neither a or b has any elements

In [7]: if not (a.size and b.size):
   ...:     print('neither a or b has any elements')
   ...:     

In [8]: if not (a and b):
   ...:     print('neither a or b has any elements')
   ...:     
neither a or b has any elements

In [9]: if not (np.size(a) and np.size(b)):
   ...:     print('neither a or b has any elements')
   ...:     

In [10]: 

In [16]: a = np.array([0]); b = np.array([0])

In [17]: print(a); print(b)
[0]
[0]

In [18]: if not a or not b:
    ...:     print('neither a or b has any elements')
    ...:     
neither a or b has any elements

In [19]: if not (a and b):
    ...:     print('neither a or b has any elements')
    ...:     
neither a or b has any elements

In [20]: if not (a.size and b.size):
    ...:     print('neither a or b has any elements')
    ...:     

In [21]: if not (np.size(a) and np.size(b)):
    ...:     print('neither a or b has any elements')
    ...:  
```

jhamman: nice. yes, `a.size` is still an improvement.

nmizukami:
Hi @jhamman  so here are remaining items we might want to improve, but I probably want to close this for now.  I think this code improved substantially compared to the beginning (in terms of performance and  readability) .    1) xarray reading vs netcdf4 reading.  2) refactoring split function  3) arguments.

jhamman:
sure, just open separate issues for those 3 and we can return to them later. You still need to resolve the merge conflicts here before I can merge this.

nmizukami:
@jhamman   I think i resolved the conflicts ?

jhamman:
@nmizukami  - not yet. Still Conflicts.

nmizukami:
I think now it is fixed

jhamman:
please open follow up issues for the remaining tasks

PR #34, opened 2018-02-20
Feature/update lohmann

martynpclark:
Mostly modularization


route/build/src/read_remap.f90
```diff
@@ -4,101 +4,115 @@ module read_remap
...
-use globalData,  only:remap_data
+use read_netcdf, only:get_nc_dim_len
+use read_netcdf, only:get_units
 
 implicit none
 
 private
 
 public::get_remap_data
+public::get_runoff_metadata
```
nmizukami: This routine seems to be out of place. This routine is for runoff data not for remap. maybe move this to read_runoff.f90?

martynpclark:
@nmizukami -- I addressed your review comment.

The refactor here is now almost complete. We just need to add the routing code, and these subroutines do not need major changes.

PR #38, opened 2018-03-05
Feature/add seg2seg irf

nmizukami:
Modified IRF routing module
1. included IRF routing driver ( this intends to be called from main routing program) called "irf_route".
2. segment loop is removed from convolution routine. I intend to loop through segment in irf_route routine. 
3. Removed commented lines.
4. Some editing comments.

### DIFF
--- a/route/build/src/dataTypes.f90
+++ b/route/build/src/dataTypes.f90
@@ -103,14 +103,14 @@ implicit none
   real(DP)                                :: RCHLAT2  ! End latitude
   real(DP)                                :: RCHLON1  ! Start longitude
   real(DP)                                :: RCHLON2  ! End longitude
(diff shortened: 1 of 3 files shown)


nmizukami:
@martynpclark  Added back IRF routing with minimum change in route_runoff.f90.

martynpclark:
@nmizukami -- can you summarize the tests that you did?

nmizukami:
@martynpclark. I tested with https://github.com/nmizukami/mizuRoute/blob/mod_irf_test. This includes the same make_uh and conv_upsbas_qr routines in irf_route.f90 as the one I pull requested. also call convolution routine the same way in route_runoff.f90.     I ran this with my Cameo test case (/glade/p/work/mizukami/test_mizuRoute/water_balance_check - where I am testing uh).    I think I wanted to test NHDPlus2 with SUMMA

PR #53, opened 2018-09-20
fix bugs in the gridded runoff capabilities

martynpclark:


route/build/src/remap.f90
```diff
@@ -151,6 +163,8 @@ subroutine remap_2D_runoff(runoff_data, remap_data, structHRU2seg, basinRunoff,
 
   end do   ! looping through basins in the routing layer
```
nmizukami: This comment and the line 288 should be 
> looping through basins in the mapping layer

route/build/src/remap.f90
```diff
@@ -113,6 +113,18 @@ subroutine remap_2D_runoff(runoff_data, remap_data, structHRU2seg, basinRunoff,
     jj = remap_data%j_index(ixOverlap)
     ii = remap_data%i_index(ixOverlap)
 
+    ! check i-indices
```
nmizukami: This is good catch.  Maybe more descriptive comments help?  
something like...

> 'When computing runoff at ', jHRU, 'th HRU, No valid runoff value is found at i index =', ii, 'in runoff data.'

route/build/src/remap.f90
martynpclark: Thanks @nmizukami -- can you implement this change?

martynpclark:
Thanks @nmizukami -- can you implement these changes?

PR #68, opened 2019-07-23
Update process_param.f90

ShervanGharari:
The change takes care of the case if the shape function and time scale are less than the modelling time step. if the UH are less than the modelling time step they set the UH length to 1. This was creating problems for the previous implementation. I have tested it with my set up; for daily it set UH length to 1 and for hourly simulation it set it to 4 or 5 hours which is actually what is should do...

### DIFF
--- a/route/build/src/process_param.f90
+++ b/route/build/src/process_param.f90
@@ -56,23 +56,33 @@ contains
   ! ---------------------------------------------------------------------------------------
   ! initialize error control
   ierr=0; message='basinUH/'
-  ! use a Gamma distribution with shape parameter, fshape = 2.5, and time parameter, tscale, input
+  ! use a Gamma distribution with shape parameter, fshape, and time parameter, tscale, input
   alamb = fshape/tscale                  ! scale parameter
   ! find the desired number of future time steps
-  ntdh_min = 1._dp
-  ntdh_max = 1000._dp
-  ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)
-  do itry=1,maxtry
-   x_value = alamb*dt*ntdh_try
-   cumprob = gammp(fshape, x_value)
-   !print*, tscale, ntdh_try, cumprob
-   if(cumprob < 0.99_dp)  ntdh_min = ntdh_try
-   if(cumprob > 0.999_dp) ntdh_max = ntdh_try
-   if(cumprob > 0.99_dp .and. cumprob < 0.999_dp) exit
-   ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)
-   if(itry==maxtry)then; ierr=20; message=trim(message)//'cannot identify the maximum number of bins for the tdh'; return; endif
-  end do
+  ! in case check if the cummulative Gamma distribution is 1 for given model time step, tscale and fsahpe.
+  ! Modified by Shervan Gharari on 23th July 2019
+  X_VALUE = alamb*dt                        ! assuming ntdh_try is set to 1.00
+  cumprob = gammp(fshape, X_VALUE)          ! check the cumdist gamma distribution for the X_VALUE of UH of 1 time step
+  if(cumprob > 0.999_dp) then               ! check if the cum dist function reaches close to one for UH of 1 time step
+   print*, cumprob, X_VALUE                 ! print the value of X_VALUE
+   ntdh_try = 1.5                           ! to make sure that we have 2.00 time step for UH one will be with [1.00,0.00]
+  else                                      ! if the sum is not one then find the best UH length by itterating from a large number
+   ntdh_min = 1._dp
+   ntdh_max = 1000._dp
+   ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)  
+   do itry=1,maxtry
+    x_value = alamb*dt*ntdh_try
+    cumprob = gammp(fshape, x_value)
+    print*, tscale, ntdh_try, cumprob, x_value, itry
+    if(cumprob < 0.99_dp)  ntdh_min = ntdh_try
+    if(cumprob > 0.999_dp) ntdh_max = ntdh_try
+    if(cumprob > 0.99_dp .and. cumprob < 0.999_dp) exit
+    ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)
+    if(itry==maxtry)then; ierr=20; message=trim(message)//'cannot identify the maximum number of bins for the tdh'; return; endif
+   end do
+  endif                                    ! end the if statement
   ntdh = ceiling(ntdh_try)
+  print*, ntdh                             ! print the ntdh
   ! allocate space for the time-delay histogram
   if (.not.allocated(FRAC_FUTURE)) then
     allocate(FRAC_FUTURE(ntdh), stat=ierr)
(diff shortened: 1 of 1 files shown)


route/build/src/process_param.f90
```diff
@@ -56,23 +56,33 @@ SUBROUTINE basinUH(dt, fshape, tscale, IERR, MESSAGE)
...
-   cumprob = gammp(fshape, x_value)
-   !print*, tscale, ntdh_try, cumprob
-   if(cumprob < 0.99_dp)  ntdh_min = ntdh_try
-   if(cumprob > 0.999_dp) ntdh_max = ntdh_try
-   if(cumprob > 0.99_dp .and. cumprob < 0.999_dp) exit
-   ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)
-   if(itry==maxtry)then; ierr=20; message=trim(message)//'cannot identify the maximum number of bins for the tdh'; return; endif
-  end do
+  ! in case check if the cummulative Gamma distribution is 1 for given model time step, tscale and fsahpe.
+  ! Modified by Shervan Gharari on 23th July 2019
```
martynpclark: You should not leave names embedded in the code. This gets very messy very quickly. GitHub tracks who did what when.

route/build/src/process_param.f90
```diff
@@ -56,23 +56,33 @@ SUBROUTINE basinUH(dt, fshape, tscale, IERR, MESSAGE)
...
+  X_VALUE = alamb*dt                        ! assuming ntdh_try is set to 1.00
+  cumprob = gammp(fshape, X_VALUE)          ! check the cumdist gamma distribution for the X_VALUE of UH of 1 time step
+  if(cumprob > 0.999_dp) then               ! check if the cum dist function reaches close to one for UH of 1 time step
+   print*, cumprob, X_VALUE                 ! print the value of X_VALUE
+   ntdh_try = 1.5                           ! to make sure that we have 2.00 time step for UH one will be with [1.00,0.00]
```
martynpclark: I think it would be clearer to have `ntdh_try = 1.9999_dp` -- note that we specify the precision of real numbers

route/build/src/process_param.f90
```diff
@@ -56,23 +56,33 @@ SUBROUTINE basinUH(dt, fshape, tscale, IERR, MESSAGE)
...
-   if(cumprob > 0.99_dp .and. cumprob < 0.999_dp) exit
-   ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)
-   if(itry==maxtry)then; ierr=20; message=trim(message)//'cannot identify the maximum number of bins for the tdh'; return; endif
-  end do
+  ! in case check if the cummulative Gamma distribution is 1 for given model time step, tscale and fsahpe.
+  ! Modified by Shervan Gharari on 23th July 2019
+  X_VALUE = alamb*dt                        ! assuming ntdh_try is set to 1.00
+  cumprob = gammp(fshape, X_VALUE)          ! check the cumdist gamma distribution for the X_VALUE of UH of 1 time step
+  if(cumprob > 0.999_dp) then               ! check if the cum dist function reaches close to one for UH of 1 time step
+   print*, cumprob, X_VALUE                 ! print the value of X_VALUE
```
martynpclark: You should not have print statements in the operational code. It should be removed or commented out.

route/build/src/process_param.f90
```diff
@@ -56,23 +56,33 @@ SUBROUTINE basinUH(dt, fshape, tscale, IERR, MESSAGE)
...
+   ntdh_try = 1.5                           ! to make sure that we have 2.00 time step for UH one will be with [1.00,0.00]
+  else                                      ! if the sum is not one then find the best UH length by itterating from a large number
+   ntdh_min = 1._dp
+   ntdh_max = 1000._dp
+   ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)  
+   do itry=1,maxtry
+    x_value = alamb*dt*ntdh_try
+    cumprob = gammp(fshape, x_value)
+    print*, tscale, ntdh_try, cumprob, x_value, itry
```
martynpclark: See above. Remove or comment out print statements.

route/build/src/process_param.f90
```diff
@@ -56,23 +56,33 @@ SUBROUTINE basinUH(dt, fshape, tscale, IERR, MESSAGE)
...
+    if(cumprob < 0.99_dp)  ntdh_min = ntdh_try
+    if(cumprob > 0.999_dp) ntdh_max = ntdh_try
+    if(cumprob > 0.99_dp .and. cumprob < 0.999_dp) exit
+    ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)
+    if(itry==maxtry)then; ierr=20; message=trim(message)//'cannot identify the maximum number of bins for the tdh'; return; endif
+   end do
+  endif                                    ! end the if statement
   ntdh = ceiling(ntdh_try)
+  print*, ntdh                             ! print the ntdh
```
martynpclark: See above. Remove or comment out print statements.

route/build/src/process_param.f90
ShervanGharari: right! got it...

route/build/src/process_param.f90
ShervanGharari: were commented out...

route/build/src/process_param.f90
ShervanGharari: done! I learn about this precision yesterday... yes it makes more sense this way...

route/build/src/process_param.f90
ShervanGharari: commented out...

route/build/src/process_param.f90
ShervanGharari: commented out

route/build/src/process_param.f90
```diff
@@ -56,23 +56,33 @@ SUBROUTINE basinUH(dt, fshape, tscale, IERR, MESSAGE)
...
-   x_value = alamb*dt*ntdh_try
-   cumprob = gammp(fshape, x_value)
-   !print*, tscale, ntdh_try, cumprob
-   if(cumprob < 0.99_dp)  ntdh_min = ntdh_try
-   if(cumprob > 0.999_dp) ntdh_max = ntdh_try
-   if(cumprob > 0.99_dp .and. cumprob < 0.999_dp) exit
-   ntdh_try = 0.5_dp*(ntdh_min + ntdh_max)
-   if(itry==maxtry)then; ierr=20; message=trim(message)//'cannot identify the maximum number of bins for the tdh'; return; endif
-  end do
+  ! in case check if the cummulative Gamma distribution is 1 for given model time step, tscale and fsahpe.
```
nmizukami: Just suggestion:  Change the comment to "check if the cumulative Gamma distribution is 1 for one future time step, given tscale and fsahpe. if so ntdh will be set to 2"

route/build/src/process_param.f90
nmizukami: For this section, I would remove comments for each line

route/build/src/process_param.f90
ShervanGharari: good suggestion... if I understand you correctly in that case first all the iteration are done and then the length is set to 1 or 2... is that what you mean?

nmizukami: your edit is fine with me.  I just wanted to clarify the purpose of if-statement you added.

PR #78, opened 2020-01-16
Feature/mpi pio

nmizukami:
this is still working in progress.


nmizukami:
Major accomplishments

- Now restart I/O working properly. 

- Now routing using single node working (compilation with MPISERIAL is not working yet)

- Now all routing computations (mainstem and tributary routing) use mainstem only domain (master node) and tributary domains (master node + other nodes). Previously mainstem routing use global domain data. 

- Clean up data structure

- output file name convention

- There are CTSM coupling stuff (cpl directory and cime_config), but can be ignored for stand-alone.

PR #89, opened 2020-03-25
added a lake_route and put some new data strcuture on the dataTypes.f…

ShervanGharari:
…90, irf_route is slightly chnaged,
I have compiled the code and it works for HDMA_global...


nmizukami:
It looks like this is suggested changes and is not meant merged to a base branch (as good working version).  So not trying to pull request and I (and maybe Martyn) can go to directly your feature/mpi-pio.lake branch and take a look.

PR #90, opened 2020-03-26
Feature/mpi pio.lake

ShervanGharari:
A lake_route.f90 file is added. And it is called inside the irf_route.f90.
The source code is successfully compiled on Graham and the result is similar. I would suggest you also test this on your test cases.


martynpclark:
Can you provide a more detailed explanation of what you have done (just a list of bullets), what tests you have completed, and what you would like in code review (e.g., general structure, detailed line-by-line comments)?

ShervanGharari:
> Can you provide a more detailed explanation of what you have done (just a list of bullets), what tests you have completed, and what you would like in code review (e.g., general structure, detailed line-by-line comments)?

The general idea is to pass the information using the RCHTOPO to account for lake in a simple fashion not to include additional data structure as these additional data structure should be propagated in all the subroutines and functions

The changes are as follow:
1- creation of lake_route.f90 that provide a simple level pool routing (linear reservoir at this point)
2- change in dataType.f90 to provide a lakeflage ISLAKE in RCHTOPO data type; at this moment set to .false. as there is not lake
3- change in irf_route.f90 to conditionally run the irf routing scheme or lake.
4- makefile should be change to include lake_route (it is not in this pull) as I didn't not commit the local makefile.

I will comment each line using git hub commenting tool

nmizukami:
> 4- makefile should be change to include lake_route (it is not in this pull) as I didn't not commit the local makefile.
yes, you don't have to track your local makefiles, but I think you still want to update Makefile (the one tracked) and include in this pull request. i believe you need to only add lake_route.f90 to list of F90 files in Makefile.

ShervanGharari:
Few extra changes
1- the code can now read <varname_islake>; it should be tested more if this reading is actually happening by maybe printing the values passed to lake_route.f90
2- there was an issue with the timing of the volume; now fixed
3- the code complies and seems to give identical result with earlier source code.

ShervanGharari:
test
removed the hardcoded .false. values in the dataType.f90 for islake.
tested for test case with lakes, it seems the current code cannot allocate true for lake segment (it is always false).

nmizukami:
> tested for test case with lakes, it seems the current code cannot allocate true for lake segment (it is always false).

I suspect you missed this one in process_ntopo.f90?
`NETOPO_in(iSeg)%isLake(iSeg) = (structNTOPO(iSeg)%var(ixNTOPO%isLake)%dat(1)==true)`

ShervanGharari:
> > tested for test case with lakes, it seems the current code cannot allocate true for lake segment (it is always false).
> 
> I suspect you missed this one in process_ntopo.f90?
> `NETOPO_in(iSeg)%isLake(iSeg) = (structNTOPO(iSeg)%var(ixNTOPO%isLake)%dat(1)==true)`

yes, indeed I missed this as I misunderstood my own notes!

I think the code should be sth link this maybe:
NETOPO_in(iSeg)%isLake = structNTOPO(iSeg)%var(ixNTOPO%isLake)%dat(1)

added at line 472 (https://github.com/NCAR/mizuRoute/blob/feature/mpi-pio/route/build/src/process_ntopo.f90#L472)

the code compiled but returns an error when reading the islake before starting the simulation:

```
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable DAT when it is not allocated

Image              PC                Routine            Line        Source
route_runoff.mpi-  00000000008D3E30  Unknown               Unknown  Unknown
route_runoff.mpi-  00000000005A1D41  process_ntopo_mp_         474  process_ntopo.f90
route_runoff.mpi-  00000000006A0C4B  mpi_routine_mp_co         342  mpi_process.f90
route_runoff.mpi-  000000000072C544  init_model_data_m         212  init_model_data.f90
route_runoff.mpi-  00000000007375FB  model_setup_mp_in          82  model_setup.f90
route_runoff.mpi-  0000000000740B95  MAIN__                     81  route_runoff.f90
route_runoff.mpi-  0000000000410E0E  Unknown               Unknown  Unknown
libc-2.24.so       00002B77D27882E0  __libc_start_main     Unknown  Unknown
route_runoff.mpi-  0000000000410D2A  Unknown               Unknown  Unknown
forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable DAT when it is not allocated```

I assume sth wrong with dat? does it say sth

File: route/build/src/popMetadat.f90 [hunk 1]
```diff
@@ -136,6 +136,7 @@ subroutine popMetadat(err,message)
  meta_NTOPO  (ixNTOPO%isLakeInlet    ) = var_info('isLakeInlet'    , 'flag to define if a lake inlet (1=true)'            ,'-'    ,ixDims%seg   , .false.)
  meta_NTOPO  (ixNTOPO%userTake       ) = var_info('userTake'       , 'flag to define if user takes water (1=true)'        ,'-'    ,ixDims%seg   , .false.)
  meta_NTOPO  (ixNTOPO%goodBasin      ) = var_info('goodBasin'      , 'flag to define a good basin (1=true)'               ,'-'    ,ixDims%upSeg , .false.)
+ meta_NTOPO  (ixNTOPO%islake         ) = var_info('islake'         , 'flag to define if the object is lake (1=true)'      ,'-'    ,ixDims%upSeg , .false.) ! has added the lake flag
```
nmizukami: Ok,  `ixDims%upSeg` is not correct.    it should be `ixDims%seg`.   isLake is not information for upstream reach, but for current reach.

ShervanGharari: compiles and runs but still does not enter to the lake routing as if all the segments are false...

ShervanGharari:
The lake output is saved in the netcdf file with variable called "IRFlakeVol".
for the lake volume to be saved I have changed the following:
1- popmetadata line 152
2- varlookup 139 and 172
3-write_simoutput 139

following is three lake in my river network. the volume is in m3. it is Bow river basin and therefore the maximum storage is happening at the end of melt period in June/July (~day 200ish)

lake_vol

next is to read lake parameters and pass it to the lake_route.f90

route/build/src/irf_route.f90
```diff
@@ -105,7 +110,14 @@ subroutine irf_route(iEns,          &  ! input: index of runoff ensemble to be p
      seg:do iSeg=1,river_basin(ix)%branch(iTrib)%nRch
        jSeg = river_basin(ix)%branch(iTrib)%segIndex(iSeg)
        if (.not. doRoute(jSeg)) cycle
-       call segment_irf(iEns, jSeg, ixDesire, NETOPO_IN, RCHFLX_out, ierr, cmessage)
+        if (NETOPO_in(jseg)%islake) then
+         print*, 'islake = ', NETOPO_in(jseg)%islake ! to check if it is lake
+         print*, 'reach id that is lake = ', NETOPO_in(jseg)%REACHID ! to check the reach id of lake
+         call lake_route(iEns, jSeg, ixDesire, NETOPO_in, RPARAM_in, RCHFLX_out, ierr, message)
```
nmizukami: Shervan,  you will need to add RPARAM_in in OMP specification - `!$OMP   shared(RPARAM_in)`.  See before tributary loop.

ShervanGharari:
The RATECVA and RATECVB originally under LAKPRP in dataType are moved under RCHPRP.
accordingly, dataType.f90, popmetadat.f90, varjookup.f90, read_control.f90, process_ntopo.f90 and mpi_process.f90 are changed. Moreover, to pass the reach parameters to lake_route.f90, main_route.90, irf_route.f90 and lake_route.f90 are changed.
the source code is compiled and the RATECVA and RATECVB are passed to lake_route (printed)

route/build/src/irf_route.f90
ShervanGharari: added, I still need to check if it compiles. also we add the job with more node (currently only one node).

nmizukami: yes, it would be good to run with several nodes at this point.  if you do that, make sure that changing number of node (also called MPI tasks) won't change the results.  nccmp command can be used to compare two netcdfs.   Note that changing number of node will change reach and hru orders in output (this is related to PIO).  I can send you small python script to reorder one netcdf to match with order of another.

ShervanGharari: would be awesome to have your python code to order the segments. I have nccmp for comparison.

ShervanGharari: the current code (which changed today to read RATECVA and B) results in the same lake volume as from from the changes before. I can do a more detailed checked for each reach as well.
lake_vol_new_3node

nmizukami: oh, because you have not used rating curve parameters in lake yet?

ShervanGharari: yes, exactly. it just make sure that the code runs and give the similar result. the parameters are passed to the subroutine. The next steps are:
1- using the parameters for stage discharge relationship (including making sure storage is not negative, etc etc)
2- add the in-basin routing output to the lake (maybe we should skip the in-basin routing when lake flag is on)

ShervanGharari:
the implementation of storage release is done using the two parameters RATECVA and RATECVB; in this implementation the parameters are assume to be unit-less [-] and therefore the A*S**B is divided by dt to yield discharge

ShervanGharari:
the file get_basin_runoff.f90 is changed so the it remaps runoff, evaporation and precipitation for all the basins; the file mpi_process.f90 is also changes to that it scatter the new basin fluxes, evaporation and precipitation across CPU; the code compiled; the remap is also change but remapping of the 3D varibale should be checked closely as it may have issues in it; the evaporation precipitation are scattered and are taken in the main_route and also lake function. basin_UH is also modified to have a pulse response function although not needed as the lake do not input in basin routing for excluding double counting (it can be set to what it was).

ShervanGharari:
The evaporation and precipitation have are changed into m3/s similar to runoff. this will help easier manipulation inside lake_route subroutines.

route/build/src/lake_route.f90
```diff
@@ -0,0 +1,137 @@
...
+  ! no runoff input is added to the lake; the input are only precipitation and evaporation to the lake
+
+  print*, '------lake-simulation-------- '
+  print*, 'node id that is lake ...... = ', NETOPO_in(segIndex)%REACHID ! to check the reach id of lake
+  print*, 'lake param RATECVA ........ = ', RPARAM_in(segIndex)%RATECVA
+  print*, 'lake param RATECVB ........ = ', RPARAM_in(segIndex)%RATECVB
+  print*, 'volume before simulation m3 = ', RCHFLX_out(iens,segIndex)%REACH_VOL(0)
+  print*, 'upstream streamflow m3/s .. = ', RCHFLX_out(iens,segIndex)%REACH_Q_IRF
+  print*, 'upstream precipitation m3/s = ', RCHFLX_out(iens,segIndex)%basinprecip
+  print*, 'upstream evaporation m3/s . = ', RCHFLX_out(iens,segIndex)%basinevapo
```
nmizukami, line 101: Yes, i think these could remain but could use ixDesire to enable spot check at a particular lake.  ixDesire is index of reach (here lake) for which this information is printed out.

route/build/src/lake_route.f90
```diff
@@ -0,0 +1,137 @@
...
+  RCHFLX_out(iens,segIndex)%REACH_Q_IRF = RPARAM_in(segIndex)%RATECVA * (RCHFLX_out(iens,segIndex)%REACH_VOL(1) ** RPARAM_in(segIndex)%RATECVB)! simplified level pool liner reservoir Q=AS^B
+  RCHFLX_out(iens,segIndex)%REACH_Q_IRF = (min(RCHFLX_out(iens,segIndex)%REACH_Q_IRF * dt, RCHFLX_out(iens,segIndex)%REACH_VOL(1)) )/dt! in case is the output volume is more than lake volume
+  RCHFLX_out(iens,segIndex)%REACH_VOL(1) = RCHFLX_out(iens,segIndex)%REACH_VOL(1) - RCHFLX_out(iens,segIndex)%REACH_Q_IRF * dt ! updating the storage
+  if (RCHFLX_out(iens,segIndex)%REACH_VOL(1) .LT. 0) then; ! set the lake volume as 0 if it goes negative
+    RCHFLX_out(iens,segIndex)%REACH_VOL(1)=0
+  endif
+
+
+  print*, 'lake simulated output m3/s .= ', RCHFLX_out(iens,segIndex)%REACH_Q_IRF
+  print*, 'volume after simulation m3 .= ', RCHFLX_out(iens,segIndex)%REACH_VOL(1)
```
nmizukami, line 120: Also use ixDesire for spot-check?

route/build/src/lake_route.f90
```diff
@@ -0,0 +1,137 @@
...
+
+  ! calculate water balance (in this water balance we dont have the actual evaporation)
+  WB = q_upstream * dt + RCHFLX_out(iens,segIndex)%basinprecip * dt - RCHFLX_out(iens,segIndex)%REACH_Q_IRF * dt &
+  - RCHFLX_out(iens,segIndex)%basinevapo * dt - (RCHFLX_out(iens,segIndex)%REACH_VOL(1) - RCHFLX_out(iens,segIndex)%REACH_VOL(0))
+  print*, 'water balance error ....... = ', WB
```
nmizukami, line 125: For this one,   print out WB when it is above threshold as warning or even terminate  ?

route/build/src/main_route.f90
```diff
@@ -98,15 +104,45 @@ SUBROUTINE main_route(&
...
+                   NETOPO_in,          & ! input: reach topology
+                   RPARAM_in,          & ! input: reach parameter
+                   reachPrecip_local,  & ! output: reach Precip (m3/s)
+                   ierr, cmessage,     & ! output: error control
+                   ixRchProcessed)       ! optional input: indices of reach to be routed
+  if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
+
+  !print*, "basin evapo in, inside main route m3/s = ", reachRunoff_local 
+  !print*, "basin runoff in, inside main route m3/s = ", reachEvapo_local
+  !print*, "basin precip in, inside main route m3/s = ", reachPrecip_local
```
nmizukami, line 145: could remove once you are confident this works.

route/build/src/irf_route.f90
```diff
@@ -5,11 +5,14 @@ module irf_route_module
 ! data type
 USE dataTypes, ONLY: STRFLX         ! fluxes in each reach
 USE dataTypes, ONLY: RCHTOPO        ! Network topology
+USE dataTypes, ONLY: RCHPRP         ! Network parameter
```
nmizukami, line 8: for comment,   could be Network parameter -> reach/lake property parameter?

route/build/src/irf_route.f90
```diff
@@ -98,14 +103,22 @@ subroutine irf_route(iEns,          &  ! input: index of runoff ensemble to be p
...
      seg:do iSeg=1,river_basin(ix)%branch(iTrib)%nRch
        jSeg = river_basin(ix)%branch(iTrib)%segIndex(iSeg)
        if (.not. doRoute(jSeg)) cycle
-       call segment_irf(iEns, jSeg, ixDesire, NETOPO_IN, RCHFLX_out, ierr, cmessage)
+        if (NETOPO_in(jseg)%islake) then
+         ! print*, 'islake = ', NETOPO_in(jseg)%islake ! to check if it is lake
+         ! print*, 'reach id that is lake = ', NETOPO_in(jseg)%REACHID ! to check the reach id of lake
+         call lake_route(iEns, jSeg, ixDesire, NETOPO_in, RPARAM_in, RCHFLX_out, ierr, message)
+        else
+         ! print*, 'isnotlake =', NETOPO_in(jseg)%islake ! to check if it is not a lake
```
nmizukami, line 119: could remove

route/build/src/irf_route.f90
```diff
@@ -98,14 +103,22 @@ subroutine irf_route(iEns,          &  ! input: index of runoff ensemble to be p
...
 !$OMP          shared(ix, iEns, ixDesire)               & ! indices shared
 !$OMP          firstprivate(nTrib)
    trib:do iTrib = 1,nTrib
      seg:do iSeg=1,river_basin(ix)%branch(iTrib)%nRch
        jSeg = river_basin(ix)%branch(iTrib)%segIndex(iSeg)
        if (.not. doRoute(jSeg)) cycle
-       call segment_irf(iEns, jSeg, ixDesire, NETOPO_IN, RCHFLX_out, ierr, cmessage)
+        if (NETOPO_in(jseg)%islake) then
+         ! print*, 'islake = ', NETOPO_in(jseg)%islake ! to check if it is lake
+         ! print*, 'reach id that is lake = ', NETOPO_in(jseg)%REACHID ! to check the reach id of lake
```
nmizukami, line 116: could remove these now?

route/build/src/standalone/get_basin_runoff.f90
```diff
@@ -61,20 +67,62 @@ subroutine get_hru_runoff(ierr, message)     ! output: error control
...
 
-  runoff_data%basinRunoff=basinRunoff
+  runoff_data%basinRunoff = basinRunoff
+  runoff_data%basinEvapo = basinEvapo
+  runoff_data%basinPrecip = basinPrecip
+
+  !print*, basinRunoff
+  !print*, basinEvapo
+  !print*, basinPrecip
+  ! stop
```
nmizukami, line 125: could remove now?

route/build/src/standalone/read_runoff.f90
```diff
@@ -254,6 +273,42 @@ subroutine read_1D_runoff(fname,          &  ! input: filename
 
  ! reshape
  runoff_data_in%qsim(1:nSpace) = dummy(1:nSpace,1)
+ !Print*, "runoff values -> ", dummy
+
+ !Print*, "varibale name -> ", vname_evasim
+ !Print*, "file name -> ", fname
```
nmizukami, line 279: could remove now?

route/build/src/standalone/read_runoff.f90
```diff
@@ -254,6 +273,42 @@ subroutine read_1D_runoff(fname,          &  ! input: filename
...
+ ! get the _fill_values for runoff variable
+ call get_var_attr_real(trim(fname), vname_precip, '_FillValue', fill_value, ierr, cmessage)
+ if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
+
+ ! replace _fill_value with -999 for dummy
+ where ( abs(dummy - fill_value) < verySmall ) dummy = realMissing
+
+ ! reshape
+ runoff_data_in%Precip(1:nSpace) = dummy(1:nSpace,1)
+ ! Print*, "precipitation values in read_runoff.f dummy -> ", dummy
```
nmizukami, line 310: could remove now?

route/build/src/standalone/read_runoff.f90
```diff
@@ -254,6 +273,42 @@ subroutine read_1D_runoff(fname,          &  ! input: filename
...
+ call get_var_attr_real(trim(fname), vname_evasim, '_FillValue', fill_value, ierr, cmessage)
+ if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
+
+ ! replace _fill_value with -999 for dummy
+ where ( abs(dummy - fill_value) < verySmall ) dummy = realMissing
+
+ ! reshape
+ runoff_data_in%Easim(1:nSpace) = dummy(1:nSpace,1)
+ ! Print*, "evaporation values -> ", dummy
+
```
nmizukami, line 295: could remove now?

route/build/src/process_ntopo.f90
```diff
@@ -415,10 +415,14 @@ USE globalData,    ONLY: fshape, tscale     ! basin IRF routing parameters (Tran
...
-   RPARAM_in(iSeg)%R_MAN_N =     structSEG(iSeg)%var(ixSEG%man_n)%dat(1)
-   RPARAM_in(iSeg)%R_WIDTH =     structSEG(iSeg)%var(ixSEG%width)%dat(1)
+   RPARAM_in(iSeg)%RLENGTH   =     structSEG(iSeg)%var(ixSEG%length)%dat(1)
+   RPARAM_in(iSeg)%R_SLOPE   = max(structSEG(iSeg)%var(ixSEG%slope)%dat(1), min_slope)
+   RPARAM_in(iSeg)%R_MAN_N   =     structSEG(iSeg)%var(ixSEG%man_n)%dat(1)
+   RPARAM_in(iSeg)%R_WIDTH   =     structSEG(iSeg)%var(ixSEG%width)%dat(1)
+   RPARAM_in(iSeg)%RATECVA   =     structSEG(iSeg)%var(ixSEG%RATECVA)%dat(1)
+   RPARAM_in(iSeg)%RATECVB   =     structSEG(iSeg)%var(ixSEG%RATECVB)%dat(1)
+   ! print*, 'Parameter A = ', structSEG(iSeg)%var(ixSEG%RATECVA)%dat(1)
+   ! print*, 'Parameter B = ', structSEG(iSeg)%var(ixSEG%RATECVB)%dat(1)
```
nmizukami, line 425: could remove these now?

route/build/src/process_ntopo.f90
```diff
@@ -469,7 +473,16 @@ USE globalData,    ONLY: fshape, tscale     ! basin IRF routing parameters (Tran
...
-   ! NOT USED: lake parameters
+   ! lake parameters
+   ! USED
+   NETOPO_in(iSeg)%isLake = (structNTOPO(iSeg)%var(ixNTOPO%isLake)%dat(1)==true)
+   ! print*, 'is lake = ', structNTOPO(iSeg)%var(ixNTOPO%isLake)%dat(1)
+   ! print*, 'is lake logical = ', (structNTOPO(iSeg)%var(ixNTOPO%isLake)%dat(1)==1)
+   !if (NETOPO_in(iSeg)%isLake) THEN ! this IF should be cleaned
+   ! print*, 'is lake = ', NETOPO_in(iSeg)%isLake ! to check the islake flag
+   ! print*, 'is reach ID = ', NETOPO_in(iSeg)%REACHID ! to check the islake flag
+   !ENDIF
```
nmizukami, line 484: Could remove these now?

nmizukami:
also, could put your document here as a record?

route/build/src/lake_route.f90
ShervanGharari: so basically I replace the "segIndex" with "ixDesire" but what if there is not ixDesire identified in the control file...
I guess I need a logical check first... if (ixDesire/=integerMissing) then ... correct?

route/build/src/lake_route.f90
ShervanGharari: I guess I need a if statement here as well. in case ixDesire not missing then... right?

route/build/src/lake_route.f90
ShervanGharari: ok, here I can do a check with a preexisting values of minimum water balance... I would say for a lake 1 m3 is an acceptable value for each time step. the resulting WB are in order of 10e-9 (which is larger than the 10e-15 for min flow for example). for now I will try to set a value in the Global file for 10e-4 for example...

route/build/src/main_route.f90
```diff
@@ -14,7 +14,7 @@ MODULE main_route_module
 ! mapping HRU runoff to reach
 USE remapping, ONLY: basin2reach
 
-! subroutines: basin routing
+! subroutines: basin routing
```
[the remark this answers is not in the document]
ShervanGharari: I set my, Sublime text, to remove tailing spaces upon saving...

route/build/src/lake_route.f90
nmizukami: Ok,  you could follow this example (think just need if statement) https://github.com/nmizukami/mizuRoute/blob/beff9b99c304942d5077eabfc9e93a5feb647dda/route/build/src/irf_route.f90#L195
`ixDesire` is determined somewhere beginning of the program based on `desireId` in control file (optional input).

ShervanGharari: I am trying to test the changed I make here for `ixDesire` how is the control file should be to account for the desire id... it seems that is not a simple integer read...
```
! MISCELLANEOUS
case('<desireId>'   );          read(cData,*,iostat=io_error) desireId          ! turn off checks or speficy reach ID if 
```
and this is how I put it in the control file...
```
<desireId>          95                              ! turn off checks or speficy reach ID if necessary to print on screen
```

ShervanGharari: at this moment it give such a error reading the control file. I suspect I am not providing the right input...
`FATAL ERROR: init_model/read_control/problem disentangling cLines(iLine) [strin
 g=]          20`

nmizukami: Yes, looks control file correct, but based on error message, i think you have empty line in control file?    `desireId` is integer, and this is how it is read.

ShervanGharari: Yes, I had an additional like in control file. now it seems to read the ID but it still returns the -9999 in stead of the if of the `ixDesire`.

nmizukami: I think I know why.  The feature/mpi-pio you branched out does not detect ixDesire if you run in one task.     I made some changes in the other pull-request so one task runs works for this.   Do you really need to use now?   if not, you could leave it as is.   Once all are merged, hope it works.   you could try multiple task.

ShervanGharari: actually I realized this as it was not printing the ixDesire in network_topo.f90. and also when a segment is given to it instead of -9999 for seg_outlet is doesn't run... no I don't need it... I can just do a hard coded value for checking the code and add the code to the pull.

ShervanGharari: no actually I think it is not that easy to check this with hard coded values, It will propagate over all the functions and index of ID may not be the same as ID, etc... I can just add this to the pull... commented the if and we can remove the comment later when ixdesire works..

ShervanGharari:
Here is the file, I will start another pull soon on the enhancement of the reading input files.
[steps_to_implement_lake_final.docx](https://github.com/NCAR/mizuRoute/files/4513139/steps_to_implement_lake_final.docx)

PR #92, opened 2020-04-22
Feature/mpi pio.lake.forcing

ShervanGharari:
if input fluxes file name is *.txt it reads a text file and put the input file name in a dataType specified for this purpose.
if input fluxes file name is *.nc. life as usual.
next is to get the start and end of each input file, compare them with current time step and pass the right file name to read.
also the location of the changes should be spread across the code (not the optimal place to be)
Compiled successfully but not tested yet.


ShervanGharari:
so far very messy code but I understood the way to read a txt file and pass the data inside it to a structure. also the place of the changes I made should be different (should go to model_setup). next to clean take a step back and see what might be the complication of passing a txt file as fname (there are couple of dependencies of fname to be a netcdf which should be resolved.) will follow up with Naoki on that.

PR #94, opened 2020-04-28
Feature/mpi pio.lake.forcing

ShervanGharari:


ShervanGharari:
The standalone is changes so it can do the following:
1- read the *txt that contains the name of the netcdf files. (netcdf file should be in time order in *txt file, they should have the same time units e.g. "days since 2001-01-01 00:00:00" for all of them, should have the same calendar.)
2- a new subroutine in model_setup called inFile_pop, reads the txt file. get the information of every file including the calendar, time unit, time variable. it also include the iTIme of the first and last time in the netcdf file. creates a total timevar, and pass the first netcdf file as the name of file to be read. it also initialize time with a new subroutine init_time_new.
3- a new subroutine called infile_name get the input time and correct iTime and file name to be read.
4- the code compiles, the model runs, but the output files seems to have some issues (maybe not properly closed, etc) as they cannot be manipulated with cdo commands.
The general idea is that the model passes iTimelocal instead of iTime for reading the netcdf files. iTimelocal is also a global parameter similar to the iTime but updates to 1 when the simulation passes to another file.
Next is 
1- To look at the simulation and compare them with original simulation (lake without multiple input files)
2- To make sure that the output files are properly generated (e.g. closed) and area ready for further manipulation.

ShervanGharari:
it is seems the problem with output netcdf was related to the fact that I was not closing the *txt file after reading. seems to be resolved when the text file is properly closed.
the result is identical with original version (lake version) on 1 and 3 node.

ShervanGharari:
The issue with writing the restart file is fixed. (it can restart for HDMA case on Cheyenne, can restart for Bow on Graham, but does not start for HDMA on graham [which should be looked into]).
Now there are 6 parameters that are taken in for the lake/reservoir.
next, to enhance the lake representation using more parameters.

ShervanGharari:
The lake implementation has been changed to Hanasaki et al, Eq.1, the network topology including lakes are prepared...
next to properly read and write the reachvol...

ShervanGharari:
The issue with passing the islake flag to mpi_process is solved. the code compiles and produce results.

PR #112, opened 2020-06-16
Feature/mpi pio multi file

ShervanGharari:
inFile_pop is added to read the information on the multiple runoff input files and init_time is modified in standalone/model_setup.f90
The code compiles but it is not tested. Work in progress.

### DIFF
--- a/route/build/src/standalone/model_setup.f90
+++ b/route/build/src/standalone/model_setup.f90
@@ -7,7 +7,7 @@ USE public_var, ONLY: iulog              ! i/o logical unit number
 USE public_var, ONLY: debug
 USE public_var, ONLY: integerMissing
 USE public_var, ONLY: realMissing
-USE public_var, ONLY : charMissing
+USE public_var, ONLY: charMissing
 
 USE init_model_data,  ONLY: init_ntopo_data
 USE init_model_data,  ONLY: init_state_data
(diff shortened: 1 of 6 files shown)


route/build/src/standalone/model_setup.f90
```diff
@@ -127,28 +129,186 @@ SUBROUTINE init_runoff_data(pid,           & ! input: proc id
...
+               vname_time, infileinfo_data(iFile)%timeVar, 1, nTime, ierr, cmessage) ! does it needs timeVar(:)
+   print*, "time var is", infileinfo_data(iFile)%timeVar
+
+   ! get the time multiplier needed to convert time to units of days for each file
+   select case( trim( infileinfo_data(iFile)%unit(1:index(infileinfo_data(iFile)%unit,' ')) ) )
+    case('seconds'); convTime2Days=86400._dp
+    case('minutes'); convTime2Days=1440._dp
+    case('hours');   convTime2Days=24._dp
+    case('days');    convTime2Days=1._dp
+    case default;    ierr=20; message=trim(message)//'unable to identify time units'; return
```
nmizukami, line 251: can allow 'second', 'sec' and 's' so can change it to case('seconds','second','sec','s'). Similarly, 
case('minutes','minute','min'),  case('hours','hour','hr','h'), case('days','day','d')
also more meaningful error message like `'<time_units>= '//trim(t_unit)//': <time_units> must be seconds, minutes, hours or days.'` after trim(message)//

ShervanGharari:
The new changes and the original code results in similar simulation. print statements are removed.
It is good to check the code with the case in which the nc files reference date are different per file.
pull ready for merging

PR #118, opened 2020-07-13
Spatial weight data handling

nmizukami:
Issue: For some spatial weight files, if there is no hydrologic model hrus within routing hru (ovelaps=0), zero is assigned to corresponding elements in data dimension variables (weight, and overlapPolyId). This means sum of overlaps over the entire routing hrus does not match with data dimension size. Current remap procedure (remap.f90) does not increment data dimension when such routing hrus are encountered when looping through routing hru.

New check_remap_data routing checks this and remove the elements in data dimension variables corresponding to such zero overlapping-routing hrus, so that the sum of overlaps is equal to the size of data dimension.

### DIFF
--- a/route/build/src/read_remap.f90
+++ b/route/build/src/read_remap.f90
@@ -81,6 +81,95 @@ contains
   if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
  end do  ! looping through variables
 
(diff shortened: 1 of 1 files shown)


nmizukami:
Test has been performed.

PR #119, opened 2020-07-16
reading and passing the evaporation and precipitation is lake is activated in control file

ShervanGharari:
The idea to have consistency with the mizuRoute set up without lakes so the same control file can be used to run the model.


ShervanGharari:
Cleaned few of the print statements, ready to be merged.

PR #124, opened 2020-07-20
implementing lake module

ShervanGharari:
Adding lake module to the mizuRoute. Most of the changes are coming from pull #90. WORK IN PROGRESS.


ShervanGharari:
Code compiles successfully. need to test it with a setup.

route/build/src/mpi_process.f90
```diff
@@ -464,6 +506,13 @@ subroutine comm_ntopo_data(pid,                & ! input: proc id
        structNTOPO_main(nRch_mainstem+ix)%var(ixNTOPO%downSegId)%dat(1) = structNTOPO(ixx)%var(ixNTOPO%downSegId)%dat(1)
        structSEG_main  (nRch_mainstem+ix)%var(ixSEG%length)%dat(1)      = structSEG(ixx)%var(ixSEG%length)%dat(1)
        structSEG_main  (nRch_mainstem+ix)%var(ixSEG%slope)%dat(1)       = structSEG(ixx)%var(ixSEG%slope)%dat(1)
+       structNTOPO_main(nRch_mainstem+ix)%var(ixNTOPO%islake)%dat(1)    = structNTOPO(ixx)%var(ixNTOPO%islake)%dat(1)
```
ShervanGharari, line 509: does this look ok? or should be structSEG instead of structNTOPO_main?

nmizukami: structNTOPO holds integer type and structSEG holds real type. islake is integer, so this looks ok

ShervanGharari: thank you, yes, the issues was with distributing islake instead that is fixed now... so the code works and yield results as it should be... still need few adjustment to incorporate islake_sim flag into the code so it runs is lakes are set as off.

ShervanGharari:
The code compiles and simulate lakes with IRF. Similar result with lake on 1 and 16 cpus (the is_lake_sim flag should be set to T in control file `<is_lake_sim>  T`). still need to incorporate is_lake_sim flag in the code for the lakes so in case is_lake_sim is not. provided the model simulate only river segments.

nmizukami:
You can compare two simulations with nccmp (one of netcdf utilities) to check two simulations are identical.

nccmp -dsf <netcdf1> <netcdf2>

One tricky thing (for mpi mizuRoute) is if you change cpu numbers, order in the netcdf output (reach and hru) become different and need to reorder one of the netcdfs to match with the other.

ShervanGharari:
I know you had a code for rearrangement, and then comparing, but I usually do few targeted comparison between a handful of segIds, so it doesn't really care about the rearrangement of the segid in the output file...

ShervanGharari:
The code compiles and simulate lakes if the lake flag is set to true. However the code needs the lake parameters to be identified in the control file even if lake is set to false. this can be done by directing the RATECVX to length or etc

PR #129, opened 2020-07-29
abstraction/injection to river segments [WIP]

ShervanGharari:
addition of reading a second nc file containing abstraction or injection to the river network


ShervanGharari:
The standalone is changed so that it can pass two file nc file name and iTime_local related to that two nc files. the code compiles and need to be tested.

PR #144, opened 2020-08-07
Feature/mpi pio sort

ShervanGharari:
The subroutine sort_runoff is generalized and simplified with more input varibale to sort_flux.
The source code compiles but should be tested to check whether or not it produce the identical result with original source code.

### DIFF
--- a/route/build/src/remap.f90
+++ b/route/build/src/remap.f90
@@ -19,7 +19,7 @@ module remapping
 
   ! global data
   USE public_var, ONLY:iulog                  ! i/o logical unit number
-  USE public_var, ONLY:runoffMin, negRunoffTol, integerMissing
+  USE public_var, ONLY:runoffMin, negRunoffTol, integerMissing, realMissing
   USE globalData, ONLY:time_conv,length_conv  ! conversion factors
 
   implicit none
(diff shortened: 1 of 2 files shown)


route/build/src/remap.f90
```diff
@@ -260,49 +260,69 @@ subroutine remap_1D_runoff(runoff_data_in, remap_data_in, basinRunoff, ierr, mes
...
-  integer(i4b), parameter            :: ixCheck=-huge(iHRU) ! basin to check
+  integer(i4b)                         :: i,j              ! index of basin in the routing layer
 
-  ierr=0; message="sort_runoff/"
+  ierr=0; message="sort_flux/"
 
-  ! loop through hrus in the runoff layer
-  do iHRU=1,size(runoff_data_in%hru_ix)
+!  ! allocate basin_flux_sort is not allocated
+!  if (.not.allocated(basin_flux_sort)) then
```
ShervanGharari, line 290: I don't know why it complains, about this line of code, for now commend
```The argument of the ALLOCATED inquiry intrinsic function should have the ALLOCATABLE attribute.   [BASIN_FLUX_SORT]
  if (.not.allocated(basin_flux_sort)) then
---------------------^```

File: route/build/src/remap.f90 [hunk 2]
```diff
@@ -260,49 +260,69 @@ subroutine remap_1D_runoff(runoff_data_in, remap_data_in, basinRunoff, ierr, mes
...
+  real(dp)             , intent(inout) :: basin_flux_sort(:)  ! inout: sorted input states and fluxes based on ID_total_in (the same size as ID_total_in)
   ! output
-  real(dp)             , intent(out) :: basinRunoff(:)   ! basin runoff
-  integer(i4b)         , intent(out) :: ierr             ! error code
-  character(len=strLen), intent(out) :: message          ! error message
+  integer(i4b)         , intent(out)   :: ierr             ! error code
+  character(len=strLen), intent(out)   :: message          ! error message
   ! local
-  integer(i4b)                       :: iHRU,jHRU        ! index of basin in the routing layer
-  real(dp)    , parameter            :: xTol=1.e-6_dp    ! tolerance to avoid divide by zero
```
ShervanGharari, line 276: I removed xTol as later in basin2reach we check the values for negative and also agains 10^-15 as minimum runoff.

route/build/src/remap.f90
nmizukami: This means `basin_flux_sort` is not allocatable array (probably fixed dimension array?)

route/build/src/remap.f90
nmizukami: yes, this sorting array does not have to check this twice.

ShervanGharari:
For CLM/HDMA result in exact same model output.
Ready to be reviewed, tested and merged.

route/build/src/remap.f90
```diff
@@ -260,49 +260,69 @@ subroutine remap_1D_runoff(runoff_data_in, remap_data_in, basinRunoff, ierr, mes
...
 
   ! *****
-  ! public subroutine: assign runoff data in runoff layer to hru in river network layer
+  ! public subroutine: assign (sort) fluxes and states from a given ID array to another ID array
   ! ***************************************************************************************************************
-  ! case 3: hru in runoff layer is hru polygon identical to river network layer (stored in 1-dimension array)
-  subroutine sort_runoff(runoff_data_in, basinRunoff, ierr, message)
+  ! case 3: sort a given input flux and its ID based on a second given ID array,
+  ! hru in runoff layer is hru polygon identical to river network layer (stored in 1-dimension array)
+  subroutine sort_flux  (ID_in,            & ! input: the array of ids for HRU/seg (can be from 1 to nHRU or nSeg)
```
nmizukami, line 267: I would remove the part `(can be from 1 to nHRU or nSeg)` since id is not necessarily sequential number.

route/build/src/remap.f90
```diff
@@ -260,49 +260,69 @@ subroutine remap_1D_runoff(runoff_data_in, remap_data_in, basinRunoff, ierr, mes
...
+!  endif
+
+!  ! check the size of ID_in, IX_in and flux_in
+!  if ((size(ID_in)/=size(IX_in)).or.(size(IX_in)/=size(flux_in)).or.(size(flux_in)/=size(ID_in))) then
+!    ierr = 30
+!    if(ierr/=0)then; message=trim(message)//'ID_in, IX_in and flux_in not the same size'; return; endif
+!  endif
+
+  ! initializing the basin_flux_sort to zero, assuming non existing elements are all set to zero
+  basin_flux_sort = 0._dp
```
nmizukami, line 308: Thinking of sorting for only selected segID/hruID, I think this is better to be initialized as realMissing since 0._dp might be valid value.  Once it is sorted into 0 initialed array, it might cause silent, runtime logical error.

route/build/src/remap.f90
```diff
@@ -260,49 +260,69 @@ subroutine remap_1D_runoff(runoff_data_in, remap_data_in, basinRunoff, ierr, mes
...
+                         ierr, message)
   implicit none
   ! input
-  type(runoff)         , intent(in)  :: runoff_data_in   ! runoff for one time step for all HRUs
+  integer(i4b)         , intent(in)    :: ID_in(:)            ! input: the array of ids for HRU/seg (can be from 1 to nHRU or nSeg)
+  integer(i4b)         , intent(in)    :: IX_in(:)            ! input: the array of location of ids for HRU/seg in ID_total_in (can be from 1 to nHRU or nSeg)
+  real(dp)             , intent(in)    :: flux_in(:)          ! input: the array of input fluxes, should be the same size as ID_in
+  integer(i4b)         , intent(in)    :: ID_total_in(:)      ! input: the sorted array of ids for HRU/seg (should be with nHRU or nSeg)
+  ! input/output
+  real(dp)             , intent(inout) :: basin_flux_sort(:)  ! inout: sorted input states and fluxes based on ID_total_in (the same size as ID_total_in)
```
nmizukami, line 280: I would rename this to even more general or for not `flux_sort`

route/build/src/remap.f90
```diff
@@ -260,49 +260,69 @@ subroutine remap_1D_runoff(runoff_data_in, remap_data_in, basinRunoff, ierr, mes
...
+  ! public subroutine: assign (sort) fluxes and states from a given ID array to another ID array
   ! ***************************************************************************************************************
-  ! case 3: hru in runoff layer is hru polygon identical to river network layer (stored in 1-dimension array)
-  subroutine sort_runoff(runoff_data_in, basinRunoff, ierr, message)
+  ! case 3: sort a given input flux and its ID based on a second given ID array,
+  ! hru in runoff layer is hru polygon identical to river network layer (stored in 1-dimension array)
+  subroutine sort_flux  (ID_in,            & ! input: the array of ids for HRU/seg (can be from 1 to nHRU or nSeg)
+                         IX_in,            & ! input: the array of location of ids for HRU/seg (the same size as ID_in)
+                         flux_in,          & ! input: the array of input fluxes, should be the same size as ID_in
+                         ID_total_in,      & ! input: the sorted array of ids for HRU/seg (should be with nHRU or nSeg) just for check the size
```
nmizukami, line 270: this can be removed? not used.

ShervanGharari:
I have a 2D case and will give it a test also.

ShervanGharari:
I have cleaned the code. It compiles once in two attempts, I guess that means there is a dependency or loop in setting up the subroutines...
it complains from here:
```
/home/shg096/mizuRoute/route/build/src/write_restart_pio.f90(207): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [PROCESS_TIME_MODULE]
   USE process_time_module, ONLY: conv_julian2cal  ! compute data and time from julian day
-------^
/home/shg096/mizuRoute/route/build/src/write_restart_pio.f90(207): error #6580: Name in only-list does not exist or is not accessible.   [CONV_JULIAN2CAL]
   USE process_time_module, ONLY: conv_julian2cal  ! compute data and time from julian day
```

ShervanGharari:
The issue was from the makefile I had, now fixed and compiles, should test next.

ShervanGharari:
I have tested the code as follow, as the 2D case I had needed remap.
I have done the following:
1- simulate HDMA global with CLM for the original code.
2- use the variable basin runoff (which is mapped to routing hrus with original code) as input to the model.
3- run the sort exe with control file that does not have mapping and only read basin runoff from the model output from step 1.
The results are identical. I have redo step 3 with 16 cpus (similar to step 1) and with 10 cpus in which both of them result in the same segment streamflow for a handful of segments as original code in step 1.

PR #162, opened 2020-12-09
Feature/mpi pio

ShervanGharari:


ShervanGharari:
Two changes:

1- the negative input values are set to zero in the read_runoff (get_basin runoff). and then pass to be model that makes life easier to ensure the positive values

2- a bug is fixed with the i_Time_wm which in model setup in which the reference day is compared to the first day of runoff that the reference day (the reference day can be easier than the first day of runoff). We need the i_Time_wm in reference to the start of the runoff simulations even if the reference Julian day start earlier (first time variable is not zero).

route/build/src/standalone/read_runoff.f90
```diff
@@ -353,6 +353,9 @@ subroutine read_1D_runoff(fname,          &  ! input: filename
  ! replace _fill_value with -999 for dummy
  where ( abs(dummy - input_fillvalue) < verySmall ) dummy = realMissing
 
+ ! replacing the negative values
```
nmizukami, line 356: if checking negative value, why not where (dummy < 0._dp) dummy = 0._dp?

ShervanGharari: I have commented them at this moment. we do this in sort now sth like ```where(sorted_flux <0._dp) sorted_flux = 0._dp```

PR #165, opened 2020-12-12
runoff fillvalue issue

nmizukami:
This change fixed the issue below.
 
- A user was not able to overwrite runoff fillvalue if runoff netcdf has fillvalue attribute. This would be problem if runoff netcdf has wrong fillvalue; e.g., runoff netcdf is an output of xarray which puts NaN in fillvalue attribute as default. 

- Avoid checking/reading fillvalue every time step. Instead, runoff fillvalue is read and saved during model setup


ShervanGharari:
I think it is good to have develop version read fillvalue once. But perhaps implementing this for mpi is a bit difficult as there are multiple variables and hence perhaps multiple fillvalues per variables. What would be the a good way for dealing with that in mpi pio?

nmizukami:
Yes, that is right. I would say, need to check fillvalue for each variables. maybe need some refactor data structures to incorporate individual fill values, calendar, and etc.

PR #168, opened 2020-12-27
implementation of multiple lake model, their parameters and past time step simulation for lakes

ShervanGharari:


ShervanGharari:
in the read_streamSeg.f90 there is a case check is added to read the three variable of "islake", "targetvol" and "lakemodeltype" into three local array to then perform the checks.
https://github.com/ShervanGharari/mizuRoute/blob/feature/mpi-pio.lake_update_Dec/route/build/src/read_streamSeg.f90#L320
the check will be:
check 1- if a segment is lake then it should either have target volume 1 and lake model type not equal to 1 or 2 (Doll or Hanasaki) or if the target volume is 0 then the parametric should be either 1 or 2 (Doll or Hanasaki)
check 2- check for parameteric lakes either either the lake model is Doll or Hanasaki, check with global flags from control file and terminate or give warning

ShervanGharari:
ok here is a question (although obvious, I spent some hours on that!)
the varName is the name of variable in the nc file which can be different from case to case and depending on how the user specify them. therefore how I can make sure that I am getting the right name of the variable specify in the pop_matadata or var_look up to make sure that I am putting the right variable into the local array I want? how can the cases be adjusted so that is the name of variables are different then that does not affect the allocation to the local array?
https://github.com/ShervanGharari/mizuRoute/blob/feature/mpi-pio.lake_update_Dec/route/build/src/read_streamSeg.f90#L335

ShervanGharari:
> ok here is a question (although obvious, I spent some hours on that!)
> the varName is the name of variable in the nc file which can be different from case to case and depending on how the user specify them. therefore how I can make sure that I am getting the right name of the variable specify in the pop_matadata or var_look up to make sure that I am putting the right variable into the local array I want? how can the cases be adjusted so that is the name of variables are different then that does not affect the allocation to the local array?
> https://github.com/ShervanGharari/mizuRoute/blob/feature/mpi-pio.lake_update_Dec/route/build/src/read_streamSeg.f90#L335

issue is solved by comparing iVar and index for islake from ixNTOPO%islake

ShervanGharari:
The checks for wether or not the lakes model types and target volume flags for each lake do not have any conflict is added to the read_streamSeg.f90
next is to implement the array of memory of past simulation, inflow values inside lake_route.f90.

ShervanGharari:
The code is ready for review;
Doll 2003 is supported; few changes should be done in the control file to be able to run the lakes with Doll
1 - The Doll flag should be set to True in the control file otherwise the code will terminate and ask the user to set the flag to true;
2- The name of the variable to Doll are changes accordingly for the Doll implementation in control file
3- The type of the lake model should be given in an extra variable in the network topology
here is an example of the change in the control file to run the Doll lake model for all the lakes in the network topology;
```
<is_lake_sim>             T         ! flag to specify that the lake simulations are on
<lake_model_D03>          T         ! flag to specify that the Doll lake parameter exists in river network topology
<varname_D03MaxStorage>   RATECVC   ! variable maximum storage
<varname_D03Coefficient>  RATECVA   ! variable coefficient it is given in seconds 0.01/(24*3600) = (1.15740741e-7) better numerical implementation is needed to take care of the variables with dimensions time
<varname_D03Power>        RATECVB   ! variable power of the function set to 1.5
<varname_islake>          ISLAKE    ! name of variable holding the islake flag (0=reach, 1=lake)
<varname_lakeModelType>   ISLAKE    ! name of variable holding the parametric type of model (1=Doll 2003; 2=Hanasaki 2006; else = etc)
```

ShervanGharari:
Code compiles and generated identical results for lakes; it should be tested for identical case for the case without lake simulations. the code should be able to simulate the setups without lake (no change in control file).

ShervanGharari:
this branch and original branch `feature/mpi-pio` result in identical values for IRF and KWT when lake is off. I think it can be tested by you one more time and merged.

PR #182, opened 2021-04-22
Implementing Hanasaki with memory

ShervanGharari:
This is the implementation of Hanasaki by Inne. the code compiles and results in identical result with original code.
Next is to add the memory for Hanasaki input (inflow over the past n years).


nmizukami:
Hi Shervan, there are conflict to merge.  would you be able to resolve the conflicts?

ShervanGharari:
Hi Naoki, I just checked the conflict, should be easy to resolve! I'll do that...

ShervanGharari:
resolved and the code compiles successfully...

ShervanGharari:
The size is changed into shape to get the dimension of the 2D array of memory. The code compiles and run for the Hanasaki setting.
The code results identical values with the `feature/mpi-pio` and can be ingested into the main repo; however more test to check the lake module with Hanasaki and memory flag on is needed.

ShervanGharari:
The calendar is added so if the calendar is no_leap or 365 the memory is only for 28 days otherwise the memory will be for 28.25 days for month of February

PR #190, opened 2021-06-22
revise Gamma distribution in develop branch

h294liu:
The previous code ACTUALLY has an error in calculating the x_value of the Gamma distribution. The reason behind is that, when x_value is employed in "cumprob = gammp(fshape, x_value)", the gammp function requires the x_value being standardized. Given the two-parameter Gamma distribution, the time delay "dt* ntdh_try" should be standardized by "dt* ntdh_try/scale". The old mizuRoute code has an error in doing the standardization. Here this issue is fixed by removing variable "alamb" and updating variable "x_value".

Firstly, the figures below show the sensitivity of the updated Gamma distribution to the parameters GammaShape and GammaScale, respectively.
shape_sensitivity
scale_sensitivity

Secondly, the figures below show the impacts of code changes on routed runoff. This simulation is based on the Bow at Banff case study. The first figure shows the summa averageRoutedRunoff of the 1st GRU. The second figure shows the mizuRoute IRFroutedRunoff at the basin outlet. You can see that new code brings down the peak runoff. It is worth noting that this simulation runs hillslope in summa and thus investigates the code changes of summa. Given the similarity of hillslope routing in summa and mizuroute, I think the code change effects shown in summa also apply to mizuRoute.
SUMMA_output
mizuRoute_output

The last but not least, the comparison between code output and analytical results. One thing I want to highlight is that the code error exists in standardizing x_value, not calculating cumprob. Therefore, when testing the differences, we should look at the x_value changes, not cumprob correctness (cumprob will be always correct once the shape parameter and a standardized x are given). The comparison of x_value is shown in the attached Excel file. Moreover, if you have interest looking at cumprob result, the attachment also shows that the code output cumprob matches the analytical result. Hope this helps :) 
[Gamma_dist_compare.xlsx](https://github.com/NCAR/mizuRoute/files/6695846/Gamma_dist_compare.xlsx)

### DIFF
--- a/route/build/src/process_param.f90
+++ b/route/build/src/process_param.f90
@@ -35,13 +35,12 @@ contains
   IMPLICIT NONE
   ! input
   REAL(DP), INTENT(IN)                   :: dt          ! model time step
-  REAL(DP), INTENT(IN)                   :: fshape      ! shapef parameter in gamma distribution
-  REAL(DP), INTENT(IN)                   :: tscale      ! time scale parameter
+  REAL(DP), INTENT(IN)                   :: fshape      ! shape parameter in gamma distribution
+  REAL(DP), INTENT(IN)                   :: tscale      ! time scale parameter in gamma distribution
   ! output
   INTEGER(I4B), INTENT(OUT)              :: IERR        ! error code
   CHARACTER(*), INTENT(OUT)              :: MESSAGE     ! error message
   ! locals
-  REAL(DP)                               :: alamb       ! scale parameter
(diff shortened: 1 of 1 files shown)


nmizukami:
I would need to test gammp function

nmizukami:
Thanks @h294liu,   I also did some check on the function [gammp](https://github.com/NCAR/mizuRoute/blob/71f1b7725bfe783b89edfafb30dbe8790158221a/route/build/src/gamma_func.f90#L13) in gamma_func.f90 with [some test code](https://github.com/nmizukami/test_gamma_pdf).

PR #203, opened 2021-08-20
Resolve pio iosystem confusion causing PIO netcdf writing

nmizukami:
Made pio iosystem initialization optional so that mizuRoute initialize pio iosytem when running in stand-alone mode, while for coupled mode, just use pio iosystem initialized by CIME (coupler).  

This resolves issue #184


nmizukami:
fixed #184

PR #219, opened 2021-10-07
commented out 'user config' part of makefile

wknoben:
This allows variable specification outside of makefile (on the command line or in a script), e.g. as:

```
#!/bin/bash

# specify makefile variables
export F_MASTER="/path/to/mizuroute"
export EXE="mizuroute.exe"
export FC="gfortran"
export FC_EXE="gfortran"
export NCDF_PATH="/path/to/netcdf/"
export FLAGS="-O3 -ffree-line-length-none -fmax-errors=0"

# compile
make -f ${F_MASTER}/build/Makefile
```
This is useful for automating mizuRoute compiling on new infrastructure, because the necessary commands can be stored in a script without the need to manually edit the makefile each time. If the user config part is not commented out, the makefile will overwrite existing variables with empty ones. 

Usage for those who prefer to manually edit makefiles is only minimally affected, because they will know need to un-comment as well as edit the variables in the file.

### DIFF
--- a/route/build/Makefile
+++ b/route/build/Makefile
@@ -2,7 +2,7 @@
 # Makefile to compile the routing model
 #========================================================================
 #
-# Need to edit FC, FC_EXE, EXE, isOpenMP (if OMP is used), F_MASTER, NCDF_PATH
+# Need to uncomment and edit FC, FC_EXE, EXE, isOpenMP (if OMP is used), F_MASTER, NCDF_PATH
 # Feel free to modify compiler flag (FLAGS)
 # Do not leave space after your edits
 #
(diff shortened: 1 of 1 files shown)


nmizukami:
Thanks @wknoben for suggestion.  Maybe would you be able to add README file for brief instruction?

but maybe before moving forward, I would like to raise this. 

Another way to compile without editing (from @ekluzek) is https://github.com/NCAR/mizuRoute/blob/cesm-coupling/route/build/README_cheyenne_build).  basically use makefile arguments to populate variables in makefile. Not sure if this way works for you, Wouter?

Once mpi-serial works for mpi version (cesm-coupling branch) perfectly (so mpi version of code can be run just like serial version and compile w/o mpi), plan to replace current serial version with mpi version eventually.  so i think need to use either Wouter's way or Erik's way since both use different makefile (comment out empty variables in makefile, or keep variables in makefile empty).  

I don't have strong opinion, but what do you think? @wknoben and @ekluzek?    I also understand Wouter may want to compile mizuRoute in the same way as SUMMA.

PR #234, opened 2021-12-17
Refine flag control for lake simulation

ShervanGharari:
The pull request is to make the standalone so that the variables of runoff, precipitation and evaporation are not read in the case flags are set to true.


ShervanGharari:
The code compiles

PR #235, opened 2021-12-18
lake type flags to be read in read_streamSeg from river network data

ShervanGharari:
This pull request is meant to adjust the lake type flags for reading related parameters based on what is provided in the network topology. currently user should manually identify which lake types are present in network topology (and the code gracefully exists if there are discrepancies)


ShervanGharari:
capabilities are added in this PR:
1- the user does not need to set the flags for various lake types in the control file; instead the flags are automatically read from the lake type variable in the network topology
2- the abstraction and injection is cleaned up for the irf and also lake.
3- a new variable called `REACH_WM_FLUX_actual` is added to `strflx` structure. next step would be to output this variable as well if `is_wm_flux` is set to `True`
the code compile and ready to be merged after some testing.

route/build/src/read_streamSeg.f90
```diff
@@ -376,149 +494,149 @@ subroutine getData(&
...
-
-    ! check is the number of parameteric lakes and target volume sums up to the total number of lakes
-    if (number_Doll+number_Hanasaki+number_HYPE+number_TargVol == number_lakes) then
-      print*, "number of lake models and target volume models matches the total number of lakes; should be good to go!"
-    else
-      ! print*, "number of lake models and target volume models do not match the total number of lakes"
-      ierr=20; message=trim(message)//'number of lake models and target volume models do not match the total number of lakes'; return
-    endif
-  endif
+!  ! ---------- check the consistency of the lake models  ----------------------------------------------------------
```
nmizukami, line 497: These lines commented out are needed?   if not any longer, maybe remove them?

ShervanGharari: The commented block is not needed and was removed.

Issue #228, opened 2021-12-09
Known limitation, need to use the same number of MPI tasks on restart


ekluzek:
@nmizukami let me know that the same number of MPI tasks need to be used on restart as were used to create the restart file. Changing this also would take some effort. Since, this is not something normally needed we are going to leave this a known limitation.

For CESM testing this means that we expect ERP tests to fail. ERT tests should work however as they only change the number of threads.

ekluzek:
We should add a check for this when reading the restart file and gracefully fail with an error message so that this issue is obvious to the user. This will also give a easy to spot error message if you happen to run an ERP test.

ekluzek:
nNodes in globalData is the number of MPI tasks. This should be added to the restart file and then checked against nNodes to make sure they are the same. This should be added to write_restart_pio.f90. It'll need to define it and write it. On read it should read it into a temporary variable and then compare it to the global nNodes.

ekluzek:
@nmizukami if this approach sounds good I can commit it directly to the cesm-coupling branch. If you'd like to review it I can create a PR for it. Let me know which sounds best.

nmizukami:
Hi @ekluzek,  thank you for doing this. Yes, I will review it and then merge it.

ekluzek:
This is fixed for our purposes.

PR #251, opened 2022-01-16
Add number of MPI tasks to restart file to ensure they are the same when reading

ekluzek:
Add the number of MPI tasks to the restart file to ensure they are the same when starting up from a restart. This addresses #228.


nmizukami:
Looks good to me. Also tested restarting with 1) the same number of MPI tasks and different number of MPI tasks and behave as expected.   close issue #228

nmizukami:
Looks good to me. Also tested restarting with 1) the same number of MPI tasks and different number of MPI tasks and behave as expected.   close issue #228

PR #267, opened 2022-03-18
Some cesm build changes

### LINKED ISSUE #259: Change MODEL to COMP_NAME
In a recent version of cime, components need to change the use of MODEL in buildlib to COMP_NAME. I'm finding this a problem with building and running on izumi.

### LINKED ISSUE #266: Add list of files needed to be read in to mizuroute.input_data_list
In the buildnml we need to add the list of files that mizuRoute is going to read to the file: for.input_data_list.  Currently we've been relying on those files already being in place for mizuRoute to work. Adding this will allow cime to download them before the run is done. I noticed this starting to work on izumi, since it wasn't automatically downling the needed files.

ekluzek:
Bring in some important changes for the cesm build.

Issues addressed:

Fixes #259
Fixes #266


nmizukami:
Hi Erik (@ekluzek),  you pull-requested to main branch(This is still serial version main trunk). I guess you meant to merge into cesm-coupling?

ekluzek:
Good point. I'll fix that.

route/build/src/mpi_utils.f90
```diff
@@ -9,6 +9,7 @@ MODULE mpi_utils
   USE globalData, ONLY: masterproc
   USE public_var, ONLY: root
   USE public_var, ONLY: iulog
+  USE shr_sys_mod,ONLY: shr_sys_flush
```
nmizukami, line 12: Hi Erik, wondering how this works for stand-alone. does stand-alone access to shr_sys_mod?

ekluzek: We can make it available. I think the standalone build will have to change a little bit, but not too much. I plan to make a crack at that a bit later.

ekluzek:
Most of the things needed here came in. But, it looks like there might be a few useful things to preserve from here. But, there are a good number of conflicts that will take time to resolve.

PR #282, opened 2022-05-19
Fixed the restart error when single core is used.

### LINKED ISSUE #280: Restart issue for mpi-serial case detected by nag compiler
There's an issue with mpi-serial for restart that the the nag compiler found. The restart fails in mpi_restart when it gets to the "size(global_ix_comm)" intrinsic function call, because global_ix_comm has not been allocated as comm_ntopo_data is not called for the serial case.

Here's the traceback that shows that issue:

> Completion(send) valueRuntime Error: /fs/cgd/data0/erik/ctsm_mizuroute/components/mizuRoute/route/build/src/mpi_process.f90, line 679: ALLOCATABLE GLOBAL_IX_COMM is not currently allocated
> Program terminated by fatal error
> /fs/cgd/data0/erik/ctsm_mizuroute/components/mizuRoute/route/build/src/mpi_process.f90, line 679: Error occurred in MPI_PROCESS:MPI_RESTART
> /fs/cgd/data0/erik/ctsm_mizuroute/components/mizuRoute/route/build/src/init_model_data.f90, line 461: Called by INIT_MODEL_DATA:INIT_STATE_DATA
> /fs/cgd/data0/erik/ctsm_mizuroute/components/mizuRoute/route/build/cpl/RtmMod.F90, line 235: Called by RTMMOD:ROUTE_INI
> /fs/cgd/data0/erik/ctsm_mizuroute/components/mizuRoute/route/build/cpl/nuopc/rof_comp_nuopc.F90, line 521: Called by ROF_COMP_NUOPC:INITIALIZEREALIZE
> /fs/cgd/data0/erik/ctsm_mizuroute/components/cmeps/cime_config/../cesm/driver/esmApp.F90, line 128: Called by ESMAPP
> =288 tag=1

One way around it is to prevent the call to mpi_restart with the following change:

``` fortran
diff --git a/route/build/src/init_model_data.f90 b/route/build/src/init_model_data.f90
index 8d45ecd..409464e 100644
--- a/route/build/src/init_model_data.f90
+++ b/route/build/src/init_model_data.f90
@@ -457,7 +457,7 @@ SUBROUTINE init_state_data(pid, nNodes, comm, ierr, message)
       TSEC(0)=T0+dt; TSEC(1)=T1+dt
     end if
 
-    if (nNodes>0) then
+    if (nNodes>1) then
       call mpi_restart(pid, nNodes, comm, iens, ierr, cmessage)
     end if
```

That allows the case to run, but it fails the restart with the field rofImp_Forr_rofl showing different answers on restart.

>  RMS rofImp_Forr_rofl                 8.1208E+02            NORMALIZED  1.3086E+00

So possibly parts of mpi_restart should still be run for the mpi-serial case, but those parts should be extracted out and always run, while the MPI specific parts should remain in mpi_restart.

nmizukami:
Bug: when a single core is used,  state data scattering failed after restart file is read (because `global_ix_comm` is initialized only when multi-core is used).

When a single core is used, some of lines (mpi scattering, and array initialization on distributed cores) do not need to be executed.

Fix #280

### DIFF
--- a/route/build/src/mpi_process.f90
+++ b/route/build/src/mpi_process.f90
@@ -676,6 +676,8 @@ CONTAINS
       RCHFLX_main(iens, iSeg) = RCHFLX(iens, jSeg)
       RCHSTA_main(iens, iSeg) = RCHSTA(iens, jSeg)
     end do
+    ! Need to add ghost reacheas (==tributary reaches in other procs feeding into mainstem)
+    ! if not multiProc (==single proc), no ghost reach. everything is mainstem.
     if (multiProcs) then
       do iSeg = 1,size(global_ix_comm)
         jSeg = global_ix_comm(iSeg)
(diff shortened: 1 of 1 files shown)


ekluzek (changes requested):
The changes are straightforward and it's nice to see you could do this with a very small change.

Two things I'd like you to do though.

First I would add a comment before the first "if (MutliPRocs)" statement to 
say why this is only done for the non-serial case.

The other suggestion is that you change the "if (not multiProcs) return" statement
into a long if statement with the rest of the subroutine. Obviously, functionally right
now it's equivalent to what you have now.  The reason I recommend the long if is that it makes it more clear about what's going on. It's really easy to miss these early return statements. So it helps with readability. But, it also can help down the line as the code is changed if something is added that should be done for the serial case but is after the return statement.  Because it's easy to miss the return statement it's easy for that to happen.

nmizukami:
thanks @ekluzek,  I made those two changes.

route/build/src/mpi_process.f90
```diff
@@ -676,133 +676,141 @@ subroutine mpi_restart(pid,           & ! input: proc id
...
+        do iSeg = 1, rch_per_proc(pid)
+          RCHFLX_trib(iens,iSeg)%ROUTE(idxIRF)%REACH_VOL(iTbound-1) = vol_local(iSeg)
+        enddo
+      end do
+    endif
+
+  end if ! (multProc)
 
   ! no need for the entire domain flux/state data strucure
   if (masterproc) then
```
ekluzek, line 816: So this deallocation should always be done including the serial case? Just making sure...

ekluzek (approved):

route/build/src/mpi_process.f90
nmizukami: Yes, RCHFLX and  RCHSTA are deallocated safely.

PR #288, opened 2022-06-28
Enabling water abstraction with Irrigation demand 

nmizukami:
- mapping irrigation flux imported from LND (which is in hru) to reach

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -4,31 +4,24 @@ MODULE RtmMod
 
   USE pio
   USE perf_mod
-  USE shr_pio_mod   , ONLY : shr_pio_getiotype, shr_pio_getioformat, &
-                             shr_pio_getrearranger, shr_pio_getioroot, shr_pio_getiosys
-  USE shr_kind_mod  , ONLY : r8 => shr_kind_r8, CL => SHR_KIND_CL
-  USE shr_sys_mod   , ONLY : shr_sys_flush, shr_sys_abort
-  USE RtmVar        , ONLY : nt_rtm, rtm_tracers, &
-                             ice_runoff, do_rtm, do_rtmflood, &
-                             nsrContinue, nsrBranch, nsrStartup, nsrest, &
-                             cfile_name, coupling_period, &
-                             caseid, brnch_retain_casename, inst_suffix, inst_name, &
-                             barrier_timers
-  USE RtmFileUtils,   ONLY : relavu, getavu, opnfil, getfil
-  USE RunoffMod     , ONLY : rtmCTL, RunoffInit
-  USE public_var,     ONLY : dt                         ! routing time step
-  USE public_var    , ONLY : iulog
-  USE public_var    , ONLY : rpntfil
-  USE globalData    , ONLY : runMode
-  USE globalData    , ONLY : iam        => pid
-  USE globalData    , ONLY : npes       => nNodes
-  USE globalData    , ONLY : mpicom_rof => mpicom_route
-  USE globalData    , ONLY : masterproc
-  USE globalData    , ONLY : multiProcs
-  USE globalData    , ONLY : pio_netcdf_format, pio_typename, pio_rearranger, &
-                             pio_root, pio_stride
-  USE globalData    , ONLY : pioSystem
-  USE nr_utility_module, ONLY : arth
+  USE shr_pio_mod,  ONLY: shr_pio_getiotype, shr_pio_getioformat, &
+                          shr_pio_getrearranger, shr_pio_getioroot, shr_pio_getiosys
+  USE shr_kind_mod, ONLY: r8 => shr_kind_r8, CL => SHR_KIND_CL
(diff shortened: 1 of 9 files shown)


route/build/cpl/RtmMod.F90
```diff
@@ -292,59 +295,74 @@ SUBROUTINE route_run(rstwr)
...
+    !----------------------------------------------------------
+    ! Mapping CLM imported fluxes to mizuRoute HRUs or Reaches
+    !-----------------------------------------------------------
+    ! --- Mapping irrigation demand [mm/s] at HRU to river reach
     ! Must be calculated before volr is updated to be consistent with lnd
-    ! Just consider land points and only remove liquid water
-    !-----------------------------------
+    call t_startf('mizuRoute_mapping_irrig')
+
+    allocate(array_temp(rtmCTL%lnumr))
```
ekluzek, line 320: As we talked about last week. Most of the allocates here, should be move to route_ini. Then you won't have to check if they are allocated. It's a good practice to have a clean method to deallocate the data as well, but currently there isn't such a method for this module (something like a route_fini subroutine). It's OK to not have a deallocate if the memory is held for the entire length of the program.

route/build/cpl/RtmMod.F90
```diff
@@ -429,19 +448,29 @@ SUBROUTINE route_run(rstwr)
...
+    rtmCTL%discharge = 0._r8
+    rtmCTL%volr      = 0._r8
+    rtmCTL%flood     = 0._r8
+
     ! put reach flux variables to associated HRUs
     if (multiProcs) then
       if (masterproc) then
         nRch_trib=rch_per_proc(0)
-        if (nRch_mainstem/=0) then
+        if (nRch_mainstem > 0) then
```
ekluzek, line 466: I'm glad you are switching the exact match for an inequality which is more robust.

route/settings/SAMPLE-coupled.control
```diff
@@ -13,8 +13,9 @@
 ! RUN CONTROL 
 ! --------------------------------------------
 <case_name>             CASE_NAME                  ! name of simulation
-<route_opt>             1                          ! option for routing schemes 0-> both, 1->IRF, 2->KWT otherwise error 
+<route_opt>             1                          ! option for routing schemes 0->accum 1->IRF (default) 2->KWT 3->KW 4->MC 5->DW
 <doesBasinRoute>        1                          ! basin routing options   0-> no, 1->IRF, otherwise error
+<is_flux_wm>            T                          ! switch for water abstraction/injection
```
ekluzek, line 18: I'm not quite sure what "is_flux_wm" does? And it looks like right now it's not connected in right? I think both the name and comment description could be improved.

route/build/cpl/RtmMod.F90
```diff
@@ -429,19 +448,29 @@ SUBROUTINE route_run(rstwr)
...
 
+    call t_stopf('mizuRoute_subcycling')
+
+    !-----------------------------------
     ! getting ready for exporting
+    !-----------------------------------
+    call t_startf('mizuRoute_prep_export')
+
+    rtmCTL%discharge = 0._r8
+    rtmCTL%volr      = 0._r8
```
ekluzek, line 459: It would really be great to see volr get set to real values, rather than just initialized to zero.

ekluzek (changes requested):
A few changes, mostly that we talked about last week. Allocation should be done in route_ini rather than route_run. volr should be set.  is_flux_wm needs a better name and description, and to be connected inside the code.

route/settings/SAMPLE-coupled.control
nmizukami: Yes, this naming is coming from different context.  Shervan was thinking about water management (wm), but we come up with more intuitive name.

route/build/cpl/RtmMod.F90
nmizukami: So here just initialization. These are supposed to be updated in line 540-542 (yes, still I need to set real volume there).  Question is do I need to initialize zero for these at each time step?

ekluzek: I wouldn't set to zero each time step, there's no reason to do that. Really you just need the correct value by the time it's sent to the mediator. It's not bad for it to be the previous value before that. It's good to have it set to zero at initialization to make sure it's set everywhere, in case later on not all cells end up being set or something. So explicitly setting it to zero each time step is a waste of CPU time. You might zero something each time step if it's a flux that's going to accumulate over different processes, but outside of that and for most state variables you needn't reinitialize to zero.

route/build/cpl/RtmMod.F90
nmizukami: Now memory allocated in preparation step and so no need to check if they are allocated.  commit [e7a5570](https://github.com/ESCOMP/mizuRoute/pull/288/commits/e7a557076c24caba84f003e4dc3ee4d2fa615376)

route/settings/SAMPLE-coupled.control
nmizukami: I will rename is_flux_wm (and also other similar one) in different pull request

PR #310, opened 2022-08-31
enable water take and injection for other routing option

nmizukami:
Other miscellaneous fixes:

use `yearly`, `monthly`, `daily` for both history and restart write frequency


nmizukami:
pull-request to wrong branch

PR #313, opened 2022-09-21
groundwork on direct_to_outlet option for negative flow handling

nmizukami:
Added new river topology variables: destination reach: `destSegId` and `destSegIndex`. Each reach points to the other reach (one-to-one link) through reach ID or index. Destination reach can be used to directly send negative runoff to the ocean for cesm-coupling or discharge transfer from one reach to another.

A routine to compute each reach to its outlet reach. 

new derived data type: `commLink`. This is used to specify links of reaches using index and processor (task)

### DIFF
--- a/route/build/src/var_lookup.f90
+++ b/route/build/src/var_lookup.f90
@@ -13,256 +13,266 @@ MODULE var_lookup
  ! ** define data structures
  ! ***********************************************************************************************************
  type, public  ::  iLook_struct
-  integer(i4b)     :: HRU          = integerMissing   ! HRU structure
-  integer(i4b)     :: HRU2SEG      = integerMissing   ! HRU-to-segment mapping structure
-  integer(i4b)     :: SEG          = integerMissing   ! stream segment structure
-  integer(i4b)     :: NTOPO        = integerMissing   ! network topology structure
-  integer(i4b)     :: PFAF         = integerMissing   ! Pfafstetter code structure
+  integer(i4b)     :: HRU          = integerMissing   ! 1. HRU structure
+  integer(i4b)     :: HRU2SEG      = integerMissing   ! 2. HRU-to-segment mapping structure
+  integer(i4b)     :: SEG          = integerMissing   ! 3. stream segment structure
+  integer(i4b)     :: NTOPO        = integerMissing   ! 4. network topology structure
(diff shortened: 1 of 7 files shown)


route/build/src/var_lookup.f90
```diff
@@ -13,256 +13,266 @@ MODULE var_lookup
...
-  integer(i4b)     :: H06_I_mem_F      = integerMissing  ! Hanasaki 2006 parameter
-  integer(i4b)     :: H06_D_mem_F      = integerMissing  ! Hanasaki 2006 parameter
-  integer(i4b)     :: H06_I_mem_L      = integerMissing  ! Hanasaki 2006 parameter
-  integer(i4b)     :: H06_D_mem_L      = integerMissing  ! Hanasaki 2006 parameter
-
+  integer(i4b)     :: basUnderLake     = integerMissing  ! 11. Area of basin under lake (m2)
+  integer(i4b)     :: rchUnderLake     = integerMissing  ! 12. Length of reach under lake (m)
+  ! Doll 2003 parameter (Natural lake outflow)
+  integer(i4b)     :: D03_MaxStorage   = integerMissing  ! 13. Lake maximum volume (m3)
+  integer(i4b)     :: D03_Coefficient  = integerMissing  ! 14.
```
ekluzek, line 91: At some point the description of these variables needs a better description of what they each are.

ekluzek (approved):
Skimming over this looks good. My one comment is that the description of the variables needs a better description for each variable. It's OK for now, but really needs a better description at some point. The general description is that they are Hanaski 2006 parameters, which is a good general statement, but some details for each one needs to be given. Going through that reference to get a bit on each particular one is important to do at some point.

route/build/src/var_lookup.f90
ShervanGharari: @ekluzek I fully agree. I will come back to this in October and I'll populate the explanation in var_lookup.f90
Just to say that the Hanasaki parameters in mizuRoute might be a different from original formulation as we have added memory as well. Also we have introduce some parameters, such as power, etc, which was hard coded in original Hanasaki formulation and that explains the large number of parameter in comparison to original Hanasaki.

ekluzek: Sounds great @ShervanGharari. Thanks for adding that update. I look forward to your working on this.

PR #315, opened 2022-09-28
Finishing up direct_to_outlet implementation

nmizukami:
- Getting destination reach index (i.e, outlet reach index) and task for each reach
- Put them in data-structure: `commRch`. 
- Add mpi one-by-one send-receive routine (mpi_utils.f90). 
- Implemented direct_to_outlet procedures in `RtmMod/run_route`.

Caveats
- procedures of transferring negative lateral flow in each to its outlet (RtmMod/run_route) requires lots of communications between tasks, and now slow. Need improvement in the future.

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -30,7 +30,7 @@ MODULE RtmMod
   USE mpi_utils, ONLY: shr_mpi_barrier
 
   implicit none
-  logical, parameter :: check=.false.
+  logical, parameter :: verbose=.false.
 
   private
   public route_ini          ! Initialize mizuRoute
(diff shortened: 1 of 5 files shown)


route/build/cpl/RunoffMod.F90
```diff
@@ -20,14 +20,14 @@ MODULE RunoffMod
...
-    real(r8), pointer :: area(:)          ! area of catchment [m2]
+    real(r8), pointer :: area(:)          ! area of hru (AKA catchment) [m2]
     integer , pointer :: mask(:)          ! reach category. 1=non outle, 2=interior basin outlet, 3=ocean outlet
 
     real(r8), pointer :: qsur(:,:)        ! coupler importing surface forcing [mm/s]
     real(r8), pointer :: qsub(:,:)        ! coupler importing subsurface forcing [mm/s]
     real(r8), pointer :: qgwl(:,:)        ! coupler importing glacier/wetland/lake forcing [mm/s]
     real(r8), pointer :: qirrig(:)        ! coupler importing irrigation [mm/s] - negative
-    real(r8), pointer :: qirrig_actual(:) ! actual irrigation and available main channel storage [mm/s]
+    real(r8), pointer :: take(:)          ! actual water take from river reach [mm/s]
```
ekluzek, line 30: I actually think the previous name qirrig_actual is a better descriptor over "take". Take could make sense if it were the sum of several fields that take water from the river.

ekluzek (changes requested):
Really glad to see this method completed. Good to know it's in place. I have a couple suggestions. The check flag could either use some comments to clarify that it's adding more verbose log output, or a different name so that it's more clear that it isn't doing additional checking. Also I think qirrig_actual is a more specific name than "take", so I'd go back to it.

route/build/cpl/RtmMod.F90
```diff
@@ -216,20 +221,20 @@ SUBROUTINE route_ini(rof_active,flood_active)
     ! index wrt global domain
     rtmCTL%gindex(rtmCTL%begr:rtmCTL%endr) = ixHRU_order(ix1:ix2)
 
-    ! hru area
+    ! additional river reach & catchment information
     if (multiProcs) then
       if (masterproc) then
         if (nRch_mainstem > 0) then
-          call get_hru_area(NETOPO_main, RPARAM_main)
+          call get_hru_area(NETOPO_main, RPARAM_main, check=check)
```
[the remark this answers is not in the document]
nmizukami: Change the optional argument to `verbose`.

route/build/cpl/RunoffMod.F90
nmizukami: rename back to `qirrig_actual`

PR #323, opened 2022-10-28
move fileVar (variable I/O control) changes in a separate suboutine

nmizukami:
Reason for this PR:

meta data for river variables, e.g., `meta_SEG`,  exist in all the processors.  One of meta data component, meta(:)%varFile can change depending on model options.   Before this PR, such changes are imbedded in `getData` routine, which is called in only main processor, so meta(:)%varFile changes happen in only main processor.  Without correct varFile in _all other processors_, the program failed somewhere Line 394-463 in mpr_process.f90. So these changes are put in a separate subroutine (`mod_meta_varFile`), which is called in all the processors.

### DIFF
--- a/route/build/Makefile
+++ b/route/build/Makefile
@@ -183,6 +183,7 @@ IO = \
 CORE = \
     accum_runoff.f90 \
     basinUH.f90 \
+    water_balance.f90 \
     lake_route.f90 \
     irf_route.f90 \
     kwt_route.f90 \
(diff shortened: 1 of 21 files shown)


nmizukami:
@ShervanGharari ,  I found change in some mpr_process.f90 done this march - #268 caused failure in distributing lake variables to the other processors.  This pull request should fix the problem.  There is no change in simulations.

PR #331, opened 2022-11-28
Doll lake formulation improvement

ShervanGharari:
- Added one parameter related to Doll lake formation (inactive lake storage) - `D03_S0`
- Enabled lake water subtraction through evaporation.  if evaporation is greater than existing volume, limit evaporation
- temporal fix on water balance check (no need to account for evaporation/precipitation for river reach)


ShervanGharari:
@nmizukami 
Hi Naoki,
the water balance check, when the lakes and reservoirs are activated, seems to be not closed and lots of warning messages are printed. The water balance in the lake module however is closed.
Also, in the current pull request, the evaporation is compared with the lake volume and adjusted in case there is not enough water in the lake to evaporation (turning potential evaporation to actual one). this variable should be inout to be used for water balance check I would say. having said that, I this also doest explain why there is so many warning generated.

nmizukami:
is it from global water balance check or per-reach or per-lake water balance?  I can imagine water balance is not closed (I believe this is from global domain water balance check). Yes, evaporation explains imbalance in water budget. 

You could turned off at least global water balance check manually (need to set `compWB` to `.false.`).

ShervanGharari:
Thank you.
I believe it is from here, which is per segment(?).
https://github.com/ESCOMP/mizuRoute/blob/cesm-coupling/route/build/src/water_balance.f90#L77
when I comment it doesn't show those many messages
I am not sure if the per segment checks should include precipitation or evaporation? or when it is routing a segment, precipitation and evaporation can be set to zero so not participating in the water balance?

nmizukami:
Yes, I did include evap and precip in river segment in water balance error check, but those are never counted in routing.   Better to set to zero for those in water balance check for now.  I don't think I set evap and precip in river segments to zero.

nmizukami:
Hi @ShervanGharari ,  is it ready for merge?   if so can you remove "draft" mode (I think just click on "Ready for review").

ShervanGharari:
> Hi @ShervanGharari , is it ready for merge? if so can you remove "draft" mode (I think just click on "Ready for review").

Hi Naoki, I might need to add few comments and some very minor changes, however they wont impact the functionality of the current pull request. This can be reviewed and merged. Thank you.

PR #335, opened 2022-12-13
Add more lake options to the cesm-coupling control file

ekluzek:
Add more of the lake options to the cesm-coupling control file. Doing a  SMS_Ld1_D.f09_f09_rHDMAlk_mg17.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default test with this passes the smoke test.


nmizukami:
looks good to me and now can run with HDMA-lake and with lake on.

PR #345, opened 2023-02-16
Add a custom grid option for cesm-coupling

### LINKED ISSUE #343: Add a custom grid option to ctsm-coupling buildnml
As currently written the buildnml will fail if you tell mizuRoute you want to run for a grid that isn't explicitly defined in the list of valid grids. This is also explicit in the python and should be moved to a data file. We need a way to set a custom grid when running in CESM.

We are proposing that if you set LND_GRID to a custom option that it would then require the user to set the specific names of the files, but that you could then create a case.

We think this should be via setting the XML varialbe

LND_GRID

to "MIZU_USRDAT" to make it similar to the CLM_USRDAT grid that is used for CTSM.

Definition of done: Implement LND_GRID=MIZU_USRDAT option

ekluzek:
Add ability to add a custom grid option (with ROF_GRID=MIZU_USRDAT) so that you can do a custom grid setup.

Fixes #343

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -52,11 +52,20 @@ def _create_control_files(case, caseroot, srcroot, confdir, inst_string, infile,
     logger.debug("  mizuRoute lnd grid is %s ", config['lnd_grid'])
     logger.debug("  mizuRoute rof grid is %s ", config['rof_grid'])
     #----------------------------------------------------
+    # Get some important values from the case
+    #----------------------------------------------------
+    rundir = case.get_value("RUNDIR")
+    dindir = case.get_value("DIN_LOC_ROOT")
+    casename = case.get_value("CASE")
+    run_type = case.get_value("RUN_TYPE")
+    run_start = case.get_value("RUN_STARTDATE")
+    #----------------------------------------------------
     # Settings that depend on the resolution
     #----------------------------------------------------
     dname_sseg = "seg"
     dname_nhru = "hru"
     is_lake_sim = "F"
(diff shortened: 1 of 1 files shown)


nmizukami:
Hi @ekluzek, thanks! but...  could you pull-request to `cesm-coupling` branch instead of `main`??

cime_config/buildnml
```diff
@@ -152,6 +152,16 @@ def _create_control_files(case, caseroot, srcroot, confdir, inst_string, infile,
        varname_segId = "seg_id"
        varname_downSegId = "Tosegment"
        varname_pfafCode = "pfaf"
+    elif (  config['rof_grid'] == "MIZU_USRDAT" ):
```
nmizukami, line 155: `ancil_dir` also needs to be set.  `ancil_dir` is directory where `fname_ntopOld` is located.   

I believe default `ancil_dir` is this. 
`<ancil_dir>                  /glade/p/cesmdata/cseg/inputdata/rof/mizuRoute/ancillary_data/     !  directory containing ancillary data (river network data)`

so add something like this?
`<ancil_dir>        "please_set_ancillary_directory_in_user_ctl_mizuroute"`

nmizukami (commented):
I noticed one thing - `ancil_dir`.   I think user needs to set this in  `user_ctl_mizuroute`

Also, when you create the case,  is a user notified that `ancil_dir` and `fname_ntopOld` need to be set?

cime_config/buildnml
ekluzek: OK, got it added in.

PR #354, opened 2023-03-10
Added mathematical temporal bias trend to adjust discharge at future time step

nmizukami:
This PR is about direct-insertion.

Before this change, simulated discharge was corrected at gauge reach based on the error computed at observation time at extended future time steps. This is called "const"

The following mathematical models are added and `QerrTrend`(integer) is used to switch the methods in a control variable:
linear (2), logistic(3), and exponential(4). const is 1.

TO-DO (future): 
Explore the possibility of moving the direct-insertion routine out of each routing routine and put it new module/subroutine to avoid repetition of the code.

Need to write RCHFLX_out%Qelapsed in restart file to properly restart during the error correction period.


nmizukami:
Some quick analysis. 

Screen Shot 2023-03-10 at 8 59 02 AM


The gauge locations on river network.

Screen Shot 2023-03-10 at 9 04 52 AM

PR #360, opened 2023-04-01
bugfixes in init and run routines for ctsm-coupling

nmizukami:
Fixing the following bug fixes

- Fixed unit conversion factor
- Fixed populating rtmCTL - area array 
- Fixed populating rtmCTL - volr array (water volume for exporting)

**These bugs caused incorrect values for one rtmCTL variable (rtmCTL%area), exporting variable (rtmCTL%volr), and many irrigation related variables, affecting water extraction from river via irrigation demand.**

Minor comments update on error handling simulation time step (<dt_qsim>), history file frequency (<newFileFrequeycn.), and history write frequency (<outputFrequency>)

Added variables for git information in globalData

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -274,19 +274,22 @@ CONTAINS
 
       SUBROUTINE get_hru_area(NETOPO_in, RPARAM_in, offset, verbose)
 
-        ! Descriptions: Compute HRU areas
-        ! Note: mizuRoute holds contributory area [m2]-BASAREA, which CAN consists of multiple HRUs
+        ! Descriptions: Get HRU areas in rtmCTL from mizuRoute NETOPO/RPARAM data structure
+        ! Note: 1) mizuRoute holds contributory area [m2]-BASAREA, which CAN consists of multiple HRUs
         !       To get HRU area associated with each reach, need to compute based on areal weight
+        !       2) offset (optional input): in main processor, index for rtmCTL variable is based on 1 through nHRU_mainstem+nHRU_trib
+        !       (combining mainstem and tributary hru in the order). Therefore, hru index based NETOPO for tributary in main processor
+        !       need to be added by number of nHRU_mainstem. offset can be used for this.
 
         USE dataTypes, ONLY: RCHTOPO   ! data structure - Network topology
         USE dataTypes, ONLY: RCHPRP    ! data structure - Reach/hru physical parameters
 
         implicit none
         ! Arguments:
-        type(RCHTOPO),     intent(in) :: NETOPO_in(:)
-        type(RCHPRP),      intent(in) :: RPARAM_in(:)
-        integer, optional, intent(in) :: offset
-        logical, optional, intent(in) :: verbose
+        type(RCHTOPO),     intent(in) :: NETOPO_in(:) ! network topology data structure
+        type(RCHPRP),      intent(in) :: RPARAM_in(:) ! reach parameter data structure
+        integer, optional, intent(in) :: offset       ! index offset to point correct location in rtmCTL
+        logical, optional, intent(in) :: verbose      ! verbose option
         ! Local variables:
         logical                   :: verb
         integer                   :: ix
@@ -714,6 +717,9 @@ CONTAINS
         ! Descriptions: get exporting variables
         !  - discharge [m3/s],
         !  - water volume per HUC area [m]
+        !  NOTE 1) offset (optional input): in main processor, index for rtmCTL variable is based on 1 through nHRU_mainstem+nHRU_trib
+        !       (combining mainstem and tributary hru in the order). Therefore, hru index based NETOPO for tributary in main processor
+        !       need to be added by number of nHRU_mainstem. offset can be used for this.
 
         USE globalData, ONLY: idxIRF   ! routing method index
         USE dataTypes, ONLY: RCHTOPO   ! data structure - Network topology
@@ -721,9 +727,9 @@ CONTAINS
 
         implicit none
         ! Arguments:
-        type(RCHTOPO), intent(in)     :: NETOPO_in(:)
-        type(STRFLX),  intent(in)     :: RCHFLX_in(:,:)
-        integer, optional, intent(in) :: offset
+        type(RCHTOPO), intent(in)     :: NETOPO_in(:)   ! mizuRoute network topology data structure
+        type(STRFLX),  intent(in)     :: RCHFLX_in(:,:) ! mizuRoute reach flux data structure
+        integer, optional, intent(in) :: offset         ! index offset to point correct location in rtmCTL
         ! Local variables:
         integer                   :: ix
         integer                   :: iens=1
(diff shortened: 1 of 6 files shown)


route/build/cpl/RtmMod.F90
```diff
@@ -283,14 +285,14 @@ SUBROUTINE get_hru_area(NETOPO_in, RPARAM_in, verbose)
         ! Arguments:
         type(RCHTOPO),     intent(in) :: NETOPO_in(:)
         type(RCHPRP),      intent(in) :: RPARAM_in(:)
+        integer, optional, intent(in) :: offset
```
ekluzek, line 288: I'd like to see a comment to explain what the purpose of offset is

route/build/cpl/RtmMod.F90
```diff
@@ -413,19 +423,19 @@ SUBROUTINE route_run(rstwr)
     rtmCTL%qirrig_actual = -1._r8* rtmCTL%qirrig  ! actual water take [mm/s] - positive (take) or negative (inject)
     do nr = rtmCTL%begr,rtmCTL%endr
       ! calculate depth of irrigation [mm] during timestep
-      irrig_depth = rtmCTL%qirrig_actual(nr)* coupling_period
+      irrig_depth = rtmCTL%qirrig_actual(nr)* delt_coupling
```
ekluzek, line 426: What is the difference between coupling_period and delt_coupling?

route/build/cpl/RtmMod.F90
```diff
@@ -638,12 +648,12 @@ SUBROUTINE route_run(rstwr)
         if (nRch_mainstem > 0) then
           lwr=1
           upr=nRch_mainstem
-          call get_river_export_data(NETOPO_main, RCHFLX_trib(:,lwr:upr))
+          call get_river_export_data(NETOPO_main(lwr:upr), RCHFLX_trib(:,lwr:upr))
```
ekluzek, line 651: Do you really need to send an array section here? If it's really the whole array there isn't a need to send the section.

route/build/cpl/RtmMod.F90
```diff
@@ -711,30 +721,23 @@ SUBROUTINE get_river_export_data(NETOPO_in, RCHFLX_in)
...
+        integer, optional, intent(in) :: offset
         ! Local variables:
         integer                   :: ix
         integer                   :: iens=1
         integer                   :: iRch, iHru
         integer                   :: nRch, nCatch
-        real(r8)                  :: vol_hru
-
-        ! reset volr to zero
-        rtmCTL%volr(:) = 0._r8
```
ekluzek, line 724: This isn't needed, so it's good to see you get rid of it.

route/build/cpl/RtmMod.F90
```diff
@@ -711,30 +721,23 @@ SUBROUTINE get_river_export_data(NETOPO_in, RCHFLX_in)
...
             ix = NETOPO_in(iRch)%HRUIX(iHru)
-
+            if (present(offset)) ix=ix+offset
             ! stream volume is split into HRUs based on HRU's areal weight for contributory area
-            vol_hru = RCHFLX_in(iens,iRch)%ROUTE(idxIRF)%REACH_VOL(1) * NETOPO_in(iRch)%HRUWGT(iHru)
-            if (vol_hru > 0._r8) then
-              rtmCTL%volr(ix) = vol_hru/rtmCTL%area(ix)
-            end if
-
+            rtmCTL%volr(ix) = RCHFLX_in(iens,iRch)%ROUTE(idxIRF)%REACH_VOL(1)*NETOPO_in(iRch)%HRUWGT(iHru)/rtmCTL%area(ix)
```
ekluzek, line 740: The difference to above is that if vol_hru is negative it won't be set to zero. So is vol_hru always zero or positive? Or is it OK to allow it to be negative?

route/build/src/read_control.f90
```diff
@@ -403,40 +403,40 @@ SUBROUTINE read_control(ctl_fname, err, message)
  ! ---------- simulation time step, output frequency, file frequency -------
  if (masterproc) then
    write(iulog,'(2a)') new_line('a'), '---- output/simulation time steps --- '
-   write(iulog,'(A,F10.1)') '  simulation frequency:     ', dt
-   write(iulog,'(2A)')      '  history file freqeuncy:   ', trim(newFileFrequency)
-   write(iulog,'(2A)')      '  history output freqeuncy: ', trim(outputFrequency)
+   write(iulog,'(A,F10.1)') '  simulation time step <dt_qsim>:             ', dt
+   write(iulog,'(2A)')      '  history file freqeuncy <newFileFrequency>:  ', trim(newFileFrequency)
+   write(iulog,'(2A)')      '  history output freqeuncy <outputFrequency>: ', trim(outputFrequency)
```
ekluzek, line 408: It's helpful to add the name of the variable in brackets. I like that.

route/build/src/read_control.f90
```diff
@@ -403,40 +403,40 @@ SUBROUTINE read_control(ctl_fname, err, message)
...
 
  ! 1. Process history output frequency
  select case(trim(outputFrequency))
    case('daily', 'monthly', 'yearly') ! do nothing
    case default
-     read(outputFrequency,'(5I)',iostat=err) nOutFreq
+     read(outputFrequency,'(I5)',iostat=err) nOutFreq
      if (err/=0) then
-       message=trim(message)//'<outputFrequency> is invalid: must be daily, monthly, yearly or integer (number of time steps)'; return
+       message=trim(message)//'problem with numeric convertin - <outputFrequency>'; return
```
ekluzek, line 417: It seems like the previous message was easier to understand?

ekluzek (commented):
Hmmm. I have some questions on this. I might be missing something. But, I ask a bunch of questions for you to think about...

route/build/src/read_control.f90
nmizukami: So this one is tricky.  if err is detected, it is from `read(outputFrequency,'(I5)',iostat=err) nOutFreq` and this is just converting outputFrequency (character) to nOutFreq (integer).   how about this?

'problem with converting <outputFrequency> to numeric'

route/build/cpl/RtmMod.F90
nmizukami: Yes, volr should be always 0 or positive (REACH_VOL is always >=0  HRUWGT>0, rtmCTL%area>0).

route/build/src/read_control.f90
ekluzek: The reason I suggest that you keep the previous one is that someone might misspell one of the valid options and this is the error that will be triggered. So the previous error is better in terms of explaining to the user what they need to do. It's more correct in terms of what went wrong -- but it's better to explain to the user what they need to do to fix the problem -- rather than be more technically correct.

route/build/cpl/RtmMod.F90
nmizukami: this is coupling_period in sec (coupling_period is given in day).

route/build/cpl/RtmMod.F90
nmizukami: The bound of NETOPO_main is 1:nRch_mainstem + nTribOutlet, and I needed to pass only 1:nRch_mainstem section for both `NETOPO_main` and `RCHFLX_trib`

route/build/cpl/RtmMod.F90
nmizukami: Check the values in NETOPO in check_river_properties in process_ntopo.f90

route/build/src/read_control.f90
nmizukami: put the original error message back as suggested.

route/build/cpl/RtmMod.F90
nmizukami: added comments on offset

PR #363, opened 2023-04-18
Write git information in history file global attributes

### LINKED ISSUE #334: Code release version
Hi  @nmizukami , 

I think it will be quite better to write the version of code while running the program. It is be quite useful to track which version the user is using. This can be added to the main_route source code from launching the program. 

Thanks, 
Ala

nmizukami:
Write out git hash, version, branch in history file. 

Misc. change.
- popMetat.f90: use mizuroute own pio variables (not calling them from pio library)

close #334

### DIFF
--- a/route/build/src/historyFile.f90
+++ b/route/build/src/historyFile.f90
@@ -16,6 +16,9 @@ MODULE historyFile
   USE globalData,        ONLY: pio_rearranger
   USE globalData,        ONLY: pio_root
   USE globalData,        ONLY: pio_stride
+  USE globalData,        ONLY: version
+  USE globalData,        ONLY: gitBranch
+  USE globalData,        ONLY: gitHash
   USE histVars_data,     ONLY: histVars
   USE pio_utils
   USE perf_mod,          ONLY: t_startf,t_stopf ! timing start/stop (GPTL library)
@@ -127,13 +130,13 @@ MODULE historyFile
 
       ! initialze iodescRchFlux and ioDescHruFlux
       call pio_decomp(this%pioSystem,    & ! input: pio system descriptor
-                      ncd_float,        & ! input: data type (pio_int, pio_real, pio_double, pio_char)
+                      ncd_float,         & ! input: data type (pio_int, pio_real, pio_double, pio_char)
                       [nRch_in],         & ! input: dimension length == global array size
                       compdof_rch,       & ! input: local->global mapping
                       this%ioDescRchFlux)
 
       call pio_decomp(this%pioSystem,    & ! input: pio system descriptor
-                      ncd_float,        & ! input: data type (pio_int, pio_real, pio_double, pio_char)
+                      ncd_float,         & ! input: data type (pio_int, pio_real, pio_double, pio_char)
                       [nHRU_in],         & ! input: dimension length == global array size
                       compdof_hru,       & ! input: local->global mapping
                       this%ioDescHruFlux)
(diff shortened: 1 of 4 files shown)


route/build/src/pio_utils.f90
```diff
@@ -503,6 +504,41 @@ END SUBROUTINE def_var
...
+    integer(i4b)     ,  intent(in)    :: varid        ! netcdf var id
+    character(len=*) ,  intent(in)    :: attname      ! netcdf attrib
+    class(*)         ,  intent(in)    :: attval       ! netcdf attrib value
+    integer(i4b)     ,  intent(out)   :: ierr         ! error code
+    character(*)     ,  intent(out)   :: message      ! error message
+
+    ierr=0; message='put_attr/'
+
+    select type (attval)
+      type is (integer(i4b))
```
ekluzek, line 526: Probably should make these the default, so it would be done the same for anything other than character.

route/build/src/historyFile.f90
```diff
@@ -309,6 +321,15 @@ SUBROUTINE createNC_rch_hru(this, nRch_in, nHRU_in, ierr, message)
         if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
       end do
 
+      call put_attr(this%pioFileDesc, ncd_global, 'mizuRoute-version', version ,ierr, cmessage)
```
ekluzek, line 324: For the CESM coupling case you should probably fill this with model_version from lnd_comp_nuopc.F90

route/build/src/historyFile.f90
```diff
@@ -309,6 +321,15 @@ SUBROUTINE createNC_rch_hru(this, nRch_in, nHRU_in, ierr, message)
...
+      if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
+
+      call put_attr(this%pioFileDesc, ncd_global, 'gitBranch', gitBranch ,ierr, cmessage)
```
ekluzek, line 327: For the coupled case, we don't have a way to fill gitBranch or gitHash.

We could maybe add this by adding it to the mizuRoute control file? Or maybe it's OK if it's empty for the CESM coupled case?

ekluzek (approved):
Had a couple comments that I ask for. Apply them as you see fit.

route/build/src/pio_utils.f90
nmizukami: Hi @ekluzek ,   Looks like the compilers do not like if these are changed to

```
 525     select type (attval)
 526       type is (character(len=*))
 527         ierr = PIO_put_att(pioFileDesc, varid, trim(attname), trim(attval))
 528       class default  ! real(dp), real(sp), integer(i4b), integer(i8b)
 529         ierr = PIO_put_att(pioFileDesc, varid, trim(attname), attval)
 530     end select
```

`error #8769: If the actual argument is unlimited polymorphic, the corresponding dummy argument must also be unlimited polymorphic.  `
 this error points to line 529.    Looking for some more information, but not really finding anything.  

I will keep this as is, and if I can find any way to change (simplify) these lines, I will change.

ekluzek: OK, it's really not that bad as it is.

The error message says that the PIO interface would have to change. So that's beyond our scope we need here.

route/build/src/historyFile.f90
nmizukami: Yes, there is a function to get a ROF model version in [rof_comp_nuopc.F90 ](https://github.com/nmizukami/mizuRoute/blob/b76659c8ee7208b8027fbe0b7ef76b5b9c35295c/route/build/cpl/nuopc/rof_comp_nuopc.F90#L401), which is copied in a ROF shared variable named `version` by calling `rofVarSet` [here](https://github.com/nmizukami/mizuRoute/blob/b76659c8ee7208b8027fbe0b7ef76b5b9c35295c/route/build/cpl/nuopc/rof_comp_nuopc.F90#L500). 

I added mizuRoute shared variable `version` in RtmVar.f90, so `model_version` is now copied directly to mizuroute shared variable `version`  in rofVarSet subroutine (RtmVar.f90)

route/build/src/historyFile.f90
nmizukami: it is ok if they are not available.

PR #364, opened 2023-04-27
reading aggregated history variables from restart netcdf

nmizukami:
This completes temporal aggregated history output capability.  This PR implements reading history variable restart so mizuRoute can restart properly (if the previous simulation ends before aggregated history variable is output)

- New type-bound procedure to read aggregated history variables from restart netcdf (in histVar_data.f90)
- Enable restarting capability for temporal aggregated history outputs.
- Fixed an error in writing aggregated history variables to restart file.
- Pio system initialization for standalone use (avoid creating separate pioSystem for each restart output, history output, reading)

### DIFF
--- a/route/build/src/histVars_data.f90
+++ b/route/build/src/histVars_data.f90
@@ -26,12 +26,8 @@ MODULE histVars_data
   USE globalData,        ONLY: pid, nNodes
   USE globalData,        ONLY: masterproc
   USE globalData,        ONLY: mpicom_route
-  USE globalData,        ONLY: pio_netcdf_format
   USE globalData,        ONLY: pio_typename
-  USE globalData,        ONLY: pio_numiotasks
-  USE globalData,        ONLY: pio_rearranger
-  USE globalData,        ONLY: pio_root
-  USE globalData,        ONLY: pio_stride
+  USE globalData,        ONLY: pioSystem
   USE nr_utils,          ONLY: arth
   USE ncio_utils,        ONLY: get_nc
   USE pio_utils
@@ -298,26 +294,16 @@ MODULE histVars_data
       logical(lgt)                       :: FileStatus            ! file open or close
       integer(i4b), allocatable          :: compdof_rch(:)        ! global reach index for PIO reading
       integer(i4b), allocatable          :: compdof_hru(:)        ! globa hru index for PIO reading
-      type(iosystem_desc_t)              :: pioSys                ! PIO I/O system data
       type(file_desc_t)                  :: pioFileDesc           ! pio file handle
       type(io_desc_t)                    :: ioDescRchHist         ! PIO domain decomposition data for reach flux [nRch]
       type(io_desc_t)                    :: ioDescHruHist         ! PIO domain decomposition data for hru runoff [nHRU]
 
       ierr=0; message='read_restart/'
 
-      ! ----------------------------------
-      ! pio initialization for restart netCDF
-      ! ----------------------------------
-      pio_numiotasks = nNodes/pio_stride
-      call pio_sys_init(pid, mpicom_route,          & ! input: MPI related parameters
-                        pio_stride, pio_numiotasks, & ! input: PIO related parameters
-                        pio_rearranger, pio_root,   & ! input: PIO related parameters
-                        pioSys)                       ! output: PIO system descriptors
-
-      call openFile(pioSys, pioFileDesc, trim(restart_name), pio_typename, ncd_nowrite, FileStatus, ierr, cmessage)
+      call openFile(pioSystem, pioFileDesc, trim(restart_name), pio_typename, ncd_nowrite, FileStatus, ierr, cmessage)
       if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
 
-      call get_nc(trim(restart_name), 'restart_time', this%timeVar, 1, ierr, cmessage) ! get restart datetime
+      call get_nc(trim(restart_name), 'history_time', this%timeVar, 1, ierr, cmessage) ! get restart datetime
       call get_nc(trim(restart_name), 'nt', this%nt, 1, ierr, cmessage)                ! get number of aggretation time steps
 
       ! get number of HRUs and Reaches for each core
@@ -359,13 +345,13 @@ MODULE histVars_data
       compdof_hru = arth(ix1, 1, ix2-ix1+1)
 
       ! initialze ioDesc
-      call pio_decomp(pioSys,            & ! input: pio system descriptor
+      call pio_decomp(pioSystem,         & ! input: pio system descriptor
                       ncd_float,         & ! input: data type (pio_int, pio_real, pio_double, pio_char)
                       [nRch],            & ! input: dimension length == global array size
                       compdof_rch,       & ! input: local->global mapping
                       ioDescRchHist)
 
-      call pio_decomp(pioSys,            & ! input: pio system descriptor
+      call pio_decomp(pioSystem,         & ! input: pio system descriptor
                       ncd_float,         & ! input: data type (pio_int, pio_real, pio_double, pio_char)
                       [nHRU],            & ! input: dimension length == global array size
                       compdof_hru,       & ! input: local->global mapping
@@ -466,9 +452,6 @@ MODULE histVars_data
       end if
 
       call closeFile(pioFileDesc, FileStatus)
-
-      call pio_sys_finalize(pioSys, ierr, cmessage)
-      if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
 
     END SUBROUTINE read_restart
 
--- a/route/build/src/histVars_data.f90
+++ b/route/build/src/histVars_data.f90
@@ -50,12 +50,12 @@ MODULE histVars_data
     integer(i4b)             :: nHru               ! number of
     integer(i4b)             :: nRch               ! number of
     real(dp)                 :: timeVar            ! time variable [unit] at nt = 1
-    real(sp), allocatable    :: basRunoff(:)       ! catchment average runoff [m/s] [nHru]
-    real(sp), allocatable    :: instRunoff(:)      ! instantaneous lateral inflow into each river/lake [m3/s]  [nRch]
-    real(sp), allocatable    :: dlayRunoff(:)      ! lateral inflow into river or lake [m3/s] for each reach [nRch]
-    real(sp), allocatable    :: discharge(:,:)     ! river/lake discharge [m3/s] for each reach/lake and routing method [nRch,nMethod]
-    real(sp), allocatable    :: elev(:,:)          ! river/lake surface water elevation [m] for each reach/lake and routing method [nRch,nMethod]
-    real(sp), allocatable    :: volume(:,:)        ! river/lake volume [m3] for each reach/lake and routing method [nRch,nMethod]
+    real(dp), allocatable    :: basRunoff(:)       ! catchment average runoff [m/s] [nHru]
+    real(dp), allocatable    :: instRunoff(:)      ! instantaneous lateral inflow into each river/lake [m3/s]  [nRch]
+    real(dp), allocatable    :: dlayRunoff(:)      ! lateral inflow into river or lake [m3/s] for each reach [nRch]
+    real(dp), allocatable    :: discharge(:,:)     ! river/lake discharge [m3/s] for each reach/lake and routing method [nRch,nMethod]
+    real(dp), allocatable    :: elev(:,:)          ! river/lake surface water elevation [m] for each reach/lake and routing method [nRch,nMethod]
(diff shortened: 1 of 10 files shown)


route/build/src/historyFile.f90
```diff
@@ -34,7 +34,7 @@ MODULE historyFile
     integer(i4b)          :: iTime=0             ! time step in output netCDF
     logical(lgt)          :: fileStatus=.false.  ! flag to indicate history output netcdf is open
     logical(lgt)          :: gageOutput=.false.  ! flag to indicate this is at-gage-only output (== output subset of reaches)
-    type(iosystem_desc_t) :: pioSystem           ! PIO system (this does not have to be initialize for each history file?)
```
nmizukami, line 37: reason why I changed `pioSystem` to `pioSys` is `pioSystem` is global variable

route/build/src/pio_utils.f90
```diff
@@ -500,6 +524,84 @@ SUBROUTINE def_var(pioFileDesc,   & ! input: file descriptor
...
+  character(*), intent(out)            :: message      ! error message
+  ! local variables
+  type(var_desc_t)                     :: pioVarId     ! netCDF variable ID
+
+  ierr=0; message='read_darray1D/'
+
+  ierr = pio_inq_varid(pioFileDesc, trim(vname), pioVarId)
+  if(ierr/=0)then; message=trim(message)//'ERROR: getting variable id'; return; endif
+
+   select type (array)
```
nmizukami, line 554: pio_read_darray does not take the input variable defined as `class(*)`

route/build/src/histVars_data.f90
```diff
@@ -148,7 +163,7 @@ SUBROUTINE aggregate(this,             & ! inout:
         write(message,'(2A,G0,A,G0)') trim(message),'history buffer reach size:',this%nRch,'/= input data reach size:',nRch_input
         ierr=81; return
       end if
-
+! need to convert double precision to single precision??
```
ekluzek, line 166: Remove this comment you just need to change the definition of the output file precision to output in single/double.

route/build/src/histVars_data.f90
```diff
@@ -125,7 +140,7 @@ SUBROUTINE aggregate(this,             & ! inout:
       integer(i4b)                       :: nHRU_input
       integer(i4b)                       :: nRch_input
       integer(i4b)                       :: ix, iRoute          ! loop indices
-      integer(i4b)                       :: idxMethod           ! temporal method index
+      integer(i4b)                       :: idxMethod             ! temporal method index
```
ekluzek, line 143: Looks like this is actually messing up the white space alignment

route/build/src/histVars_data.f90
```diff
@@ -17,6 +17,20 @@ MODULE histVars_data
...
+  USE globalData,        ONLY: rch_per_proc        ! number of reaches assigned to each proc (size = num of procs)
+  ! pio stuff
+  USE globalData,        ONLY: pid, nNodes
+  USE globalData,        ONLY: masterproc
+  USE globalData,        ONLY: mpicom_route
+  USE globalData,        ONLY: pio_typename
+  USE globalData,        ONLY: pioSystem
+  USE nr_utils,          ONLY: arth
+  USE ncio_utils,        ONLY: get_nc
+  USE pio_utils
```
ekluzek, line 33: pio routines probably start with "pio_" but it's still good to list them explicitly so that their source can be easily tracked down. You do have pio_typename and pioSystem above which could be confused with pio_utils.

route/build/src/histVars_data.f90
```diff
@@ -260,4 +275,184 @@ SUBROUTINE clean(this)
...
+        ix1 = 1
+      else
+        ix1 = sum(hru_per_proc(-1:pid-1))+1
+      endif
+      ix2 = sum(hru_per_proc(-1:pid))
+      compdof_hru = arth(ix1, 1, ix2-ix1+1)
+
+      ! initialze ioDesc
+      call pio_decomp(pioSystem,         & ! input: pio system descriptor
+                      ncd_float,         & ! input: data type (pio_int, pio_real, pio_double, pio_char)
```
ekluzek, line 349: For restarts you should ALWAYS write out data in double precision. Otherwise restarts won't be exact.

For history files you want to be able to change it in order to save disk space.

route/build/src/histVars_data.f90
```diff
@@ -260,4 +275,184 @@ SUBROUTINE clean(this)
...
+                      [nRch],            & ! input: dimension length == global array size
+                      compdof_rch,       & ! input: local->global mapping
+                      ioDescRchHist)
+
+      call pio_decomp(pioSystem,         & ! input: pio system descriptor
+                      ncd_float,         & ! input: data type (pio_int, pio_real, pio_double, pio_char)
```
ekluzek, line 355: ncd_double

route/build/src/write_restart_pio.f90
```diff
@@ -336,6 +320,9 @@ SUBROUTINE define_state_nc(fname,           &  ! input: filename
  call def_var(pioFileDescState, 'restart_time', ncd_float, ierr, cmessage, vdesc='resatart time', vunit=trim(time_units), vcal=calendar)
  if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
 
+ call def_var(pioFileDescState, 'history_time', ncd_float, ierr, cmessage, vdesc='history time', vunit=trim(time_units), vcal=calendar)
```
ekluzek, line 323: I'd make these two double for restarts. It probably won't matter, but restarts you want to ensure are exact so output in the native precision.

route/build/src/histVars_data.f90
```diff
@@ -260,4 +275,184 @@ SUBROUTINE clean(this)
...
+                      [nHRU],            & ! input: dimension length == global array size
+                      compdof_hru,       & ! input: local->global mapping
+                      ioDescHruHist)
+
+      ! reading basin runoff (hru dimension)
+      if (meta_hflx(ixHFLX%basRunoff)%varFile) then
+        allocate(this%basRunoff(this%nHru), stat=ierr, errmsg=cmessage)
+        if(ierr/=0)then; message=trim(message)//trim(cmessage)//' [hVars%basRunoff]'; return; endif
+
+        call read_pnetcdf(pioFileDesc, meta_hflx(ixHFLX%basRunoff)%varName, this%basRunoff, ioDescHruHist, ierr, cmessage)
```
ekluzek, line 365: The name read_pnetcdf makes it sound like it's using PNETCDF rather than PIO. It might be good to change the name to avoid that confusion.

route/build/src/histVars_data.f90
```diff
@@ -260,4 +275,184 @@ SUBROUTINE clean(this)
...
+        if(ierr/=0)then; message=trim(message)//trim(cmessage)//' [hVars%discharge]'; return; endif
+        allocate(this%volume(this%nRch, nRoutes), stat=ierr, errmsg=cmessage)
+        if(ierr/=0)then; message=trim(message)//trim(cmessage)//' [hVars%discharge]'; return; endif
+
+        this%discharge = 0._sp
+        this%volume    = 0._sp
+
+        do ixRoute=1,nRoutes
+          select case(routeMethods(ixRoute))
+            case(accumRunoff);           ixFlow=ixRFLX%sumUpstreamRunoff
```
ekluzek, line 415: Personally I think it's more clear to put on seperate lines rather than using ";"...

route/build/src/historyFile.f90
ekluzek: So it's being passed down by argument rather than accessing a global variable from a module? That makes sense.

I wonder about this sort of thing as well. Even though it's passed down it actually IS the global variable. So you could argue you shouldn't change the name.

Another option would be to do something like PioSystemLocal -- to note that it's a local variable. I agree I'm not sure the best way to handle this either...

route/build/src/historyFile.f90
```diff
@@ -108,7 +108,7 @@ SUBROUTINE set_compdof_rch(this, compdof_rch, nRch_in)
       integer(i4b),             intent(in)     :: nRch_in          ! total number of reaches
 
       ! initialze iodescRchFlux
-      call pio_decomp(this%pioSystem,    & ! input: pio system descriptor
+      call pio_decomp(this%pioSys,       & ! input: pio system descriptor
                       ncd_float,         & ! input: data type (pio_int, pio_real, pio_double, pio_char)
```
ekluzek, line 112: This and the next two are the parts where it would be good to be able to select if double or single precision is desired.

route/build/src/ncio_utils.f90
```diff
@@ -25,8 +25,10 @@ MODULE ncio_utils
   module procedure get_iscalar
   module procedure get_dscalar
   module procedure get_ivec
+  module procedure get_svec
```
ekluzek, line 28: I think these are maybe for different types?: integer, single-precision, and double-precision? Might be good to have a one line comment about these to say what they are for...

route/build/src/pio_utils.f90
```diff
@@ -44,6 +46,11 @@ MODULE pio_utils
   public var_desc_t
   public io_desc_t
 
+  INTERFACE read_pnetcdf
```
ekluzek, line 49: I talk about this above that pnetcdf in the name makes me think maybe you are accessing via pnetcdf directly. It might be better to name this closer to the underlying interfaces "read_array"...

route/build/src/write_simoutput_pio.f90
```diff
@@ -72,11 +72,7 @@ SUBROUTINE main_new_file(ierr, message)
       if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
 
       ! initialize and create history netcdfs
-      select case(trim(runMode))
-        case('standalone');    hist_all_network = histFile(hfileout)
-        case('cesm-coupling'); hist_all_network = histFile(hfileout, pioSys=pioSystem)
-        case default; ierr=20; message=trim(message)//'rumMode: '//trim(runMode)//': must be standalone or cesm-coupling'; return
-      end select
+      hist_all_network = histFile(hfileout, pioSys=pioSystem)
```
ekluzek, line 75: Nice consolidation to not duplicate code.

ekluzek (changes requested):
I have a few things that I suggest you do here. Some are to clarify what you are doing. I think some are more important.

route/build/src/ncio_utils.f90
nmizukami: yes, s = single precision, d=double precision, i=integer.  I agree this is not clear, but I will leave them as is for this PR.  Subsequent PR (pio_utils.f90 refactoring) will remove all of these procedures.

route/build/src/pio_utils.f90
nmizukami: changed it to `read_dist_array`

route/build/src/histVars_data.f90
nmizukami: All the shared procedures in pio_utils module are listed now

route/build/src/histVars_data.f90
nmizukami: removed.

route/build/src/historyFile.f90
nmizukami: For data within derived data type, variable name can be the same as global variable.  `histFile%pioSystem `and `pioSystem` are different. but just my preference (not use the same name). For argument variable name for procudre, they become confusing (and not sure how the compiler handle (but this is not the case here).

route/build/src/histVars_data.f90
nmizukami: fixed lineup

route/build/src/histVars_data.f90
nmizukami: changed name from `read_pnetcdf` to `read_dist_array`.

ekluzek (approved):

route/build/src/histVars_data.f90
nmizukami: type defined in pio decomposition needs to be the same as output type (so if we output history variable in single, this has to be ncd_float)

route/build/src/histVars_data.f90
nmizukami: the same logic as the previous

route/build/src/historyFile.f90
nmizukami: history file is now using single precision so this should be ncd_float

route/build/src/histVars_data.f90
nmizukami: changed it to ncd_double and write this in restart file with double precision

nmizukami:
Addressed following comments from Erik:

- write restart information in double precision, including temporal aggregated history variable.
- Use `USE` and `ONLY` statement to list all the shared routines, variables.
- Other minor clean up

PR #370, opened 2023-05-03
move git variable (version, gitBranch, gitHash) to standalone f90 

nmizukami:
**issue**
fortran file extension used is f90, which causes a compilation problem in CESM build. Root cause of the problem is likely cpp preprocessing, which require fortran file extension to be F90 (see https://github.com/j3-fortran/fortran_proposals/issues/65, or https://gcc.gnu.org/onlinedocs/gfortran/Preprocessing-Options.html). For standalone compilation, f90 extension seems to be working. 

**How it is resolved for now**
Instead of changing extension for all the files (better for consistency, and could be better?), move git variable (version, gitBranch, gitHash) assignment section to model_setup.f90, which is located under standalone directory so they are visible only for standalone compilation.

### DIFF
--- a/route/build/src/standalone/model_setup.f90
+++ b/route/build/src/standalone/model_setup.f90
@@ -63,6 +63,9 @@ CONTAINS
    USE globalData,          ONLY: pio_stride
    USE globalData,          ONLY: pioSystem
    USE globalData,          ONLY: runMode
+   USE globalData,          ONLY: version             ! mizuRoute version
+   USE globalData,          ONLY: gitBranch           ! git branch
+   USE globalData,          ONLY: gitHash             ! git commit hash
    USE mpi_process,         ONLY: pass_global_data    ! mpi globaldata copy to slave proc
    USE init_model_data,     ONLY: init_ntopo_data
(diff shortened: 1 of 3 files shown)


route/build/src/standalone/model_setup.f90
```diff
@@ -618,7 +632,7 @@ SUBROUTINE init_time(ierr, message)
   ! "Monthly" option: use 2000-01 as template calendar yr/month
   ! "Daily" option:   use 2000-01-01 as template calendar yr/month/day
   select case(lower(trim(restart_write)))
-    case('annual')
```
ekluzek, line 621: I think either is acceptable here. But, whichever one you prefer is good. Although you should also make sure the documentation and comments are consistent regarding it as well. The comment above for example is wrong unless you change it also.

ekluzek (approved):
Looks great. This should resolve the issue I saw for the build.

I do have a comment about the change of Annual to Yearly. I think either one is OK, so I'll go with your preference. But, there are at least comments that would need to change if you do change it. The control file likely would need to change also.

PR #373, opened 2023-05-04
use select type to compact the generic routine

nmizukami:
use `select type construct` as follow:

```
select type (attr_value)
   type is (integer(i4b))
      <integer type specific procedure>
   type is (integer(i8b))
      <long integer type specific procedure>
   type is (real(sp))
      <single precision float type specific procedure>
   type is (real(dp))
     <double precision float type specific procedure>
   type is (character(len=*))
      <character type specific procedure>
   class default
     ierr = 1; message=trim(message)//'ERROR: invalid array type'; return
 end select
```

This reduces several hundreds of lines

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -74,9 +74,8 @@ pygments_style = None
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-#
-#html_theme = 'alabaster'
-#html_theme = 'sphinxdoc'
+# some themes: 'alabaster'(default), 'sphinxdoc', 'sphinx_rtd_theme'
+# see https://sphinx-themes.org/ (last accessed on May 4th 2023) for more themes
 html_theme = 'sphinx_rtd_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -76,7 +76,7 @@ pygments_style = None
 # a list of builtin themes.
 #
 #html_theme = 'alabaster'
-html_theme = 'sphinxdoc'
+#html_theme = 'sphinxdoc'
 html_theme = 'sphinx_rtd_theme'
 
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -76,7 +76,7 @@
 # a list of builtin themes.
 #
 #html_theme = 'alabaster'
-html_theme = 'sphinxdoc'
+#html_theme = 'sphinxdoc'
```
ekluzek, line 79: Rather than having commented out code, maybe have a list of theme's that have been used? And some notes on them as comments?

ekluzek (approved):
This is an awesome type change to compress the code. It's hard to evaluate, but I see the principle in the change happening and it's a wonderful change to reduce duplication of code.

I have one small point about the html_theme...

docs/source/conf.py
nmizukami: I put a link to information on Sphinx themes which list all the themes.

ekluzek: I see the comment about the list of themes. The problem is that it says "see the documentation for the list", in the context of this file -- I don't know what that means. Where do I find this list? I actually looked for that list and didn't see it anywhere. Maybe it's part of the documentation on readthedocs?

Part of what I'm getting at here is a personal nitpick -- I don't like to see commented out code especially when the code is in a repository. I think there are times when it is useful to do, but mostly it's not, you can use git to query what the previous code looked like. However, there are also times in a situation like this where it might be useful to have commented out code for a couple of the valid options. However, I'd add more comments to say why you might uncomment one of the options. For example

``` fortran
#html_theme = 'sphinxdoc'    # Uncomment and use this theme is a situation like this ______ 
```

I think that kind of thing can be useful. However, most of the time I'd say you shouldn't keep commented out code in a github repository. IMHO.

nmizukami: yes, I agree with commented out codes, often distracting for me.  I further modified the comments.

Issue #372, opened 2023-05-03
Error in control file for running with CESM


ekluzek:
The list of CESM tests is failing with ctsm5.1.dev114 and hash 71d4928273dab4e841a1116da72fc70e9e54691a.

One such test is: SMS_D_Ld1.f09_f09_rHDMAlk_mg17.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default

the cesm.log is filled with this:

```
-1: 1832	 1832	   32	yes	 r10i1n33    	   32
-1: 1833	 1833	   33	yes	 r10i1n33    	   33
-1: 1834	 1834	   34	yes	 r10i1n33    	   34
-1: 1835	 1835	   35	yes	 r10i1n33    	   35
459:MPT ERROR: Rank 459(g:459) is aborting with error code 1.
459:	Process ID: 64112, Host: r10i5n2, Program: /glade/scratch/erik/SMS_D_Ld1.f09_f09_rHDMAlk_mg17.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default.20230503_105902_dikw1k/bld/cesm.exe
459:	MPT Version: HPE MPT 2.25  08/14/21 03:05:20
459:
459:MPT: --------stack traceback-------
512:MPT ERROR: Rank 512(g:512) is aborting with error code 1.
512:	Process ID: 52546, Host: r10i5n9, Program: /glade/scratch/erik/SMS_D_Ld1.f09_f09_rHDMAlk_mg17.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default.20230503_105902_dikw1k/bld/cesm.exe
512:	MPT Version: HPE MPT 2.25  08/14/21 03:05:20
512:
512:MPT: --------stack traceback-------
114:MPT ERROR: Rank 114(g:114) is aborting with error code 1.
114:	Process ID: 36340, Host: r10i1n18, Program: /glade/scratch/erik/SMS_D_Ld1.f09_f09_rHDMAlk_mg17.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default.20230503_105902_dikw1k/bld/cesm.exe
114:	MPT Version: HPE MPT 2.25  08/14/21 03:05:20
114:
114:MPT: --------stack traceback-------
299:MPT ERROR: Rank 299(g:299) is aborting with error code 1.
[...]

PR #371, opened 2023-05-03
Add some lake tests to the test list update lake dataset and fix some coupled issues

ekluzek:
Add some lake tests to the CESM test list. Also add prealpha and prebeta tests, as well as correcting some issues in the test list including the grid names.

Fixes #372
Fixes #375
Fixes #262
Fixes #238 
Fixes #281

PR #374, opened 2023-05-04
cesm-coupling bugfix

nmizukami:
Update sample control file

Partially handles  #372


ekluzek (approved):

ekluzek:
I've merged this into #371 and that branch is now working as a result (it did require some other changes as well)

PR #381, opened 2023-05-15
Fix cesm restarting: 

### LINKED ISSUE #377: Exact restart tests are failing
The following Exact restart tests are failing in comparing the restart history files to the simulation of a continuous run.

```
ERS.f09_f09_mg17.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default	  COMPARE_base_rest 		
ERS.f19_f19_mg17.I2000Clm50SpMizGs.cheyenne_gnu.mizuroute-default	  COMPARE_base_rest 		
ERS.f19_f19_mg17.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default	  COMPARE_base_rest 		
ERS.f19_f19_rHDMAlk_mg17.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default	  COMPARE_base_rest 		
ERS.nldas2_nldas2_rHDMA_mnldas2.I2000Clm50SpMizGs.cheyenne_gnu.mizuroute-default	  COMPARE_base_rest 		
ERS.nldas2_nldas2_rHDMA_mnldas2.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default	  COMPARE_base_rest 		
ERS.nldas2_nldas2_rUSGS_mnldas2.I2000Clm50SpMizGs.cheyenne_gnu.mizuroute-default	  COMPARE_base_rest 		
ERS.nldas2_nldas2_rUSGS_mnldas2.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default	  COMPARE_base_rest 
```

nmizukami:
For CESM coupling mode, restart information is not quite set (for `rtmCTL`). This affect all negative runoff handling (irrigation and qgwl). 

Initialize some `rtmCTL` variables (river volume-`rtmCTL%volr` and also river discharge-`rtmCTL%discharge`) from restart file.

Use isRestart global variable (newly created) instead of using `fname_restart_in` to detect restart run.

Fixed one history output variable (volume) initialization from restart file.

Additional fix:  Right now, precipitation and evaporation are not passed from coupler and mizuRoute assume they are zero. So need to initialize precipitation and evaporation array with zero (gnu compiler put large values). 

Fixes #377

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -27,6 +27,7 @@ MODULE RtmMod
   USE globalData,   ONLY: npes       => nNodes
   USE globalData,   ONLY: mpicom_rof => mpicom_route
   USE globalData,   ONLY: masterproc
+  USE globalData,   ONLY: isRestart                  ! restart flag
   USE mpi_utils, ONLY: shr_mpi_barrier
 
   implicit none
@@ -59,6 +60,8 @@ CONTAINS
     USE globalData,          ONLY: nHRU_mainstem               ! scalar data: number of mainstem hrus
     USE globalData,          ONLY: nHRU_trib                   ! scalar data: number of tributary hrus
     USE globalData,          ONLY: nRch_trib                   ! scalar data: number of tributary reaches
+    USE globalData,          ONLY: nTribOutlet                 ! scalar data: number of tributaries flowing to mainstem
+    USE globalData,          ONLY: RCHFLX_trib                 ! data structure: Reach flux variables (per proc, tributary)
     USE globalData,          ONLY: NETOPO_trib                 ! data structure: River Network topology (other procs, tributary)
     USE globalData,          ONLY: NETOPO_main                 ! data structure: River Network topology (main proc, mainstem)
     USE globalData,          ONLY: RPARAM_trib                 ! data structure: River parameters (other procs, tributary)
@@ -82,6 +85,7 @@ CONTAINS
     character(len=CL)          :: rof_trstr                 ! tracer string
     integer                    :: ierr                      ! error code
     integer                    :: nt, ix, ix1, ix2
+    integer                    :: lwr,upr                ! lower and upper bounds for array slicing
     character(len= 7)          :: runtyp(4)                 ! run type
     character(len=CL)          :: cmessage
     character(len=*),parameter :: subname = '(route_ini) '
@@ -249,6 +253,7 @@ CONTAINS
     if ((nsrest == nsrContinue) .or. &
         (nsrest == nsrBranch)) then
       call RtmRestGetfile()
+      isRestart=.true.
       if (nsrest == nsrContinue) then
         call init_histFile(ierr, cmessage)
         if(ierr/=0)then; call shr_sys_abort(trim(subname)//trim(cmessage)); endif
@@ -261,6 +266,24 @@ CONTAINS
 
     call init_state_data(iam, npes, mpicom_rof, ierr, cmessage)
     if(ierr/=0)then; call shr_sys_abort(trim(subname)//trim(cmessage)); endif
+
+    ! put reach flux variables to associated HRUs
+    if (isRestart) then
+      if (masterproc) then
+        if (nRch_mainstem > 0) then
+          lwr=1
+          upr=nRch_mainstem
+          call get_river_export_data(NETOPO_main(lwr:upr), RCHFLX_trib(:,lwr:upr), discharge=.false.)
+        end if
+        if (nRch_trib > 0) then
+          lwr = nRch_mainstem + nTribOutlet + 1
+          upr = nRch_mainstem + nTribOutlet + nRch_trib
+          call get_river_export_data(NETOPO_trib, RCHFLX_trib(:,lwr:upr), offset=nHRU_mainstem, discharge=.false.)
+        end if
+      else ! other processors
+        call get_river_export_data(NETOPO_trib, RCHFLX_trib, discharge=.false.)
+      end if
+    end if
 
     !-------------------------------------------------------
     ! subroutines used only route_ini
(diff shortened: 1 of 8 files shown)


ekluzek (changes requested):
To review what we just discussed. We need the if section about the isRestart flag for the control file to be done for the standalone case, but not for coupled. So coupled will set isRestart based on nsrest settings.

route/build/cpl/RtmMod.F90
```diff
@@ -249,6 +253,7 @@ SUBROUTINE route_ini(rof_active,flood_active)
     if ((nsrest == nsrContinue) .or. &
         (nsrest == nsrBranch)) then
       call RtmRestGetfile()
+      isRestart=.true.
```
ekluzek, line 256: Add a else statement to set this to false for the else condition.

route/build/src/read_control.f90
```diff
@@ -342,6 +344,13 @@ SUBROUTINE read_control(ctl_fname, err, message)
    restart_dir = output_dir
  endif
 
+ ! ---------- restart option  ---------------------------------------------------------------------
```
ekluzek, line 347: Make sure this only happens for standalone.

ekluzek: Or it could be done, but we ignore it for the coupled case?

ekluzek: This can be done by setting runMode earlier in rof_comp_nuopc, so it can be accessed at this point.

nmizukami: Added if statement to set isRestart only if `runMode` is standalone

route/build/cpl/RtmMod.F90
nmizukami: Added else statement and isRestart is set to false

route/build/cpl/RtmMod.F90
```diff
@@ -27,6 +27,7 @@ MODULE RtmMod
   USE globalData,   ONLY: npes       => nNodes
   USE globalData,   ONLY: mpicom_rof => mpicom_route
   USE globalData,   ONLY: masterproc
+  USE globalData,   ONLY: isColdStart                ! restart flag
```
ekluzek, line 30: I suggest changing the comment here...

route/build/cpl/RtmMod.F90
```diff
@@ -246,13 +247,13 @@ SUBROUTINE route_ini(rof_active,flood_active)
...
+    if ((nsrest == nsrContinue) then
       call RtmRestGetfile()
-      if (nsrest == nsrContinue) then
-        call init_histFile(ierr, cmessage)
-        if(ierr/=0)then; call shr_sys_abort(trim(subname)//trim(cmessage)); endif
-      end if
+      isColdStart=.false.
+      call init_histFile(ierr, cmessage)
+      if(ierr/=0)then; call shr_sys_abort(trim(subname)//trim(cmessage)); endif
+    else
```
ekluzek, line 255: Actually in this case now, we should listen to what's on the control file and use what it's saying...

route/build/src/init_model_data.f90
```diff
@@ -325,6 +325,7 @@ SUBROUTINE init_state_data(pid, nNodes, comm, ierr, message)
   USE globalData, ONLY: nMolecule              ! computational molecule
   USE globalData, ONLY: TSEC                   ! begining/ending of simulation time step [sec]
   USE globalData, ONLY: initHvars              ! status of history variabiable data initialization
+  USE globalData, ONLY: isColdStart            ! restart flag
```
ekluzek, line 328: Change the comment...

route/build/src/read_control.f90
```diff
@@ -28,11 +28,13 @@ SUBROUTINE read_control(ctl_fname, err, message)
  USE globalData, ONLY: meta_PFAF               ! pfafstetter code
  USE globalData, ONLY: meta_rflx               ! river flux variables
  USE globalData, ONLY: meta_hflx               ! river flux variables
+ USE globalData, ONLY: isColdStart             ! number of active routing methods
```
ekluzek, line 31: Comment needs to be updated...

ekluzek (approved):
One important change I ask for here to make sure you listen to the control file when continue_run is FALSE. And just a couple notes about updating the comments about the isColdStart flag...

Approving just to make it easier...

route/build/cpl/RtmMod.F90
nmizukami: So should I change if statement from `if (nsrest == nsrContinue) then` to `if (continue_run) then`? 

if i change now, it won't work because there is no mechanism to change continue_run in control file right now, so wait till this mechanism is implemented (in Buildnml?)

route/build/src/init_model_data.f90
nmizukami: comment updated

route/build/src/read_control.f90
nmizukami: comment updated

route/build/cpl/RtmMod.F90
nmizukami: comment updated

route/build/cpl/RtmMod.F90
ekluzek: I would leave the if statement as

`if (nsrest == nsrContinue) then`

as you point out we can't change it until the buildnml change happens. And even then you might as well leave it as above, there isn't enough of a reason to change it.

ekluzek: My point was actually to now remove the

``` fortran
else
   isColdStart = .true.
```

because now we want that to set by the control file if it isn't a nsrContinue case.

nmizukami: ok. I removed those two lines.

PR #393, opened 2023-05-29
history file name convention for gauge only output

### LINKED ISSUE #385: Rename gauge history files to *.h_gauge.*.nc
For cesm-coupling we need to rename the gauge history data files from

`<casename>_gauge.mizuroute.h.*.nc`

to

`<casename>.mizuroute.h_gauge.*.nc`

This is important to do because CESM cime archival scripts assume that ALL CESM output files are in the format:

`<casename>.mizuroute.*`

It currently can't handle files outside of that.

nmizukami:
- name convention of a history file including only gauge points.

fix #385


ekluzek (approved):
Looks good.

PR #398, opened 2023-06-03
Improving time management 

### LINKED ISSUE #392: Issue on time variable in history file for sub-daily simulation
time variable in history file get inaccurate.

**Situation 1**
mizuRoute tracks time variable (`timeVar`) with a unit and reference datetime (`time_units`) depending on a setup (can be from runoff input, or user specified). `timeVar` is written in a history file at every output frequency, and incremented by simulation time step [sec] - `dt`. if unit of time is day and `dt` is sub-daily, `timeVar` get fractions, and history file time gets slightly off actual time stamp.  This is because `dt` is divided by sec to day conversion factor (86400 sec per day) before `timeVar` incrementation.

For now, if simulation uses sub-daily, and if time unit with `hour` is provided in control file. this seems to fix this problem. 

**situation 2**
if simulation starts at some hour of day (e.g., 1980-01-01 13:00:00), `timeVar` values at the first time step gets inaccurate. This is because `timeVar` at the first time step is computed based on two Julian days (unit of day) at start datetime and reference datetime, both including fraction

nmizukami:
This corrects fractional error in time variable (`timeVar`) that is written in history and restart netCDF. This error happened when running at sub-daily e.g., 1-hour etc. and/or starting at certain time in a day. 

- Modify datatime class (datetime_data.f90). 
  - Added a few new bound-type procedures. 1. datetime delta in second (use generic operator that operates directly on datetime data. 2. compute day of year given current datetime
  - Make as many procedures pure and elemental as possible.
  - add calendar in data element

- Use datetime delta to initialize `timeVar` (time variable in netCDF) in second and increment by `dt` in second and convert it to time_unit just before writing in netCDF 

- By using datetime delta procedure, remove julian day computation, which was used to initialize `timeVar`. and adjust the arguments in datatime type-bound procedure calls.

 - use type-bound day-of-year function to compute day of year in lake_route.f90

Fix #392


ekluzek (approved):
@nmizukami and I went over this a bit and these changes all look good. I will add some comments if I notice anything, but approving.

The update to have the functions "pure elemental" is a nice improvement. And this consolidates some of the time logic and uses some standard FORTRAN time functions available, that takes care of the time management needs.

route/build/src/datetime_data.f90
```diff
@@ -147,47 +160,66 @@ SUBROUTINE sub_assign(this, that)
...
 
-  integer(i4b) FUNCTION fn_get_year(this)
+  pure elemental integer(i4b) FUNCTION fn_get_year(this)
     implicit none
     class(datetime), intent(in) :: this
     fn_get_year = this%iy
   END FUNCTION fn_get_year
 
-  integer(i4b) FUNCTION fn_get_month(this)
+  pure elemental integer(i4b) FUNCTION fn_get_month(this)
```
ekluzek, line 173: Making these pure elemental is a nice improvement.

route/build/src/datetime_data.f90
```diff
@@ -147,47 +160,66 @@ SUBROUTINE sub_assign(this, that)
...
+
+
+  integer(i4b) FUNCTION fn_dayofyear(this)
+    ! Returns the integer day of the year (ordinal date).
+    implicit none
+    class(datetime), intent(in) :: this
+    type(datetime)              :: jan1datetime
+
+    jan1datetime = datetime(this%iy, 1, 1, 0, 0, 0.0, calendar=this%calendar)
+    fn_dayofyear = int((this - jan1datetime)/86400._dp) + 1
```
ekluzek, line 217: Better to use a parameter for 86400. rather than a magic number

route/build/src/datetime_data.f90
```diff
@@ -443,4 +465,35 @@ logical(lgt) FUNCTION fn_is_le(this, that)
...
+
+    date1 = datetime(this%iy, this%im, this%id, 0, 0, 0.0, calendar=this%calendar)
+    date2 = datetime(that%iy, that%im, that%id, 0, 0, 0.0, calendar=that%calendar)
+
+    call date1%julianday(julday1, ierr, cmessage)
+    call date2%julianday(julday2, ierr, cmessage)
+
+    diffDay=julday1-julday2
+
+    ans = diffDay*86400._dp+sec_in_day1-sec_in_day2
```
ekluzek, line 495: 86400 would be better as a parameter

route/build/src/datetime_data.f90
```diff
@@ -443,4 +465,35 @@ logical(lgt) FUNCTION fn_is_le(this, that)
...
+    real(dp)                         :: sec_in_day2
+    real(dp)                         :: diffDay
+    integer(i4b)                     :: ierr
+    type(datetime)                   :: date1
+    type(datetime)                   :: date2
+    real(dp)                         :: julday1
+    real(dp)                         :: julday2
+    character(len=strLen)            :: cmessage
+
+    sec_in_day1 = real(this%ih*3600+this%imin*60, kind=dp) +this%dsec
```
ekluzek, line 484: 3600 could be a parameter for seconds in an hour.

route/build/src/standalone/model_setup.f90
```diff
@@ -609,31 +627,28 @@ SUBROUTINE init_time(ierr, message)
...
   t_unit = trim( time_units(1:index(time_units,' ')) )
   select case( trim(t_unit) )
-    case('seconds','second','sec','s'); secPerTime=1._dp; timePerDay=86400._dp
-    case('minutes','minute','min');     secPerTime=60._dp; timePerDay=1440._dp
-    case('hours','hour','hr','h');      secPerTime=3600._dp; timePerDay=24._dp
-    case('days','day','d');             secPerTime=86400._dp; timePerDay=1._dp
+    case('seconds','second','sec','s'); sec2tunit=1._dp
+    case('minutes','minute','min');     sec2tunit=60._dp
+    case('hours','hour','hr','h');      sec2tunit=3600._dp
+    case('days','day','d');             sec2tunit=86400._dp
```
ekluzek, line 644: 86400 and maybe 3600 could be parameters rather than hardcoded...

ekluzek (approved):
Just some suggestions about using magic hard coded numbers for some time values. 86400 at least would be better as a parameter rather than a hardcoded number.

route/build/src/datetime_data.f90
nmizukami: yes, there is already parameter for this (`secprday`), so replace this with the parameter

route/build/src/datetime_data.f90
nmizukami: replaced 3600 and 60 with `secprday` and `secprmin`, respectively

route/build/src/datetime_data.f90
nmizukami: replaced 86400 with secprday

PR #402, opened 2023-06-17
input time management

nmizukami:
try to use elapsed second instead of Julian day, which may cause rounding errors for sub-daily data. 

Now, we don't assume runoff time step and water-management data time step are the same.


ekluzek (approved):
Just noticed that some magic numbers could use parameters. I like the change in principle, I think you are right this will lower the problem of rounding errors.

PR #405, opened 2023-06-26
 use abstract route class

nmizukami:
Use an abstract class (`base_route`) that holds only reach routing interface and allows several specific routine methods to be instantiated by extending this class.  This way, the subroutines that loops through all the reaches  (i.e.,`accum_runoff`, `irf_route`, `kwt_route`, `kw_route`, `mc_route`, `dfw_route`), located in each routing f90, are removed and consolidated into one common routine (`route_network` in main_route.f90). 

This is just clean up and no answers should be changed.

TO-DO.   now if lake is activated (is_lake_sim is set to true) and routing methods other than IRF, the result is likely wrong. For now, if lake is activate, use only 1 for route_opt.

### DIFF
--- a/route/build/src/accum_runoff.f90
+++ b/route/build/src/accum_runoff.f90
@@ -1,147 +1,64 @@
-MODULE accum_runoff_module
+MODULE accum_route_module
 
 ! Accumulate upstream flow instantaneously
+! This is not used as routed runoff.
+! This is used to get total instantaneous upstream runoff at each reach
 
 USE nrtype
-! data type
-USE dataTypes,   ONLY: STRFLX         ! fluxes in each reach
-USE dataTypes,   ONLY: RCHTOPO        ! Network topology
-USE dataTypes,   ONLY: subbasin_omp   ! mainstem+tributary data structures
-! global data
-USE public_var,  ONLY: iulog         ! i/o logical unit number
-USE globalData,  ONLY: idxSUM        ! index of accumulation method
-! subroutines: general
-USE perf_mod,    ONLY: t_startf,t_stopf ! timing start/stop
-USE model_utils, ONLY: handle_err
+USE dataTypes,   ONLY: STRFLX          ! fluxes in each reach
+USE dataTypes,   ONLY: STRSTA          ! state in each reach
+USE dataTypes,   ONLY: RCHTOPO         ! Network topology
+USE dataTypes,   ONLY: RCHPRP          ! Reach parameter
+USE public_var,  ONLY: iulog           ! i/o logical unit number
+USE globalData,  ONLY: idxSUM          ! index of accumulation method
+USE base_route,  ONLY: base_route_rch
 
 implicit none
 
 private
-public::accum_runoff
+public::accum_route_rch
+
+type, extends(base_route_rch) :: accum_route_rch
+ CONTAINS
+   procedure, pass :: route => accum_inst_runoff
+end type accum_route_rch
 
 CONTAINS
-
- ! ---------------------------------------------------------------------------------------
- ! Public subroutine main driver for basin routing
- ! ---------------------------------------------------------------------------------------
- SUBROUTINE accum_runoff(iEns,          & ! input: index of runoff ensemble to be processed
-                         river_basin,   & ! input: river basin information (mainstem, tributary outlet etc.)
-                         ixDesire,      & ! input: ReachID to be checked by on-screen printing
-                         NETOPO_in,     & ! input: reach topology data structure
-                         RCHFLX_out,    & ! inout: reach flux data structure
-                         ierr, message, & ! output: error controls
-                         ixSubRch)        ! optional input: subset of reach indices to be processed
- ! ----------------------------------------------------------------------------------------
- ! Purpose:
- !
- ! Accumulate all the upstream delayed runoff for each reach
- ! This is not used as routed runoff.
- ! This is used to get total instantaneous upstream runoff at each reach
- !
- ! ----------------------------------------------------------------------------------------
-
- implicit none
- ! argument variables
- integer(i4b),                    intent(in)    :: iens            ! runoff ensemble index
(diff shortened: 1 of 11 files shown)


route/build/src/accum_runoff.f90
```diff
@@ -1,147 +1,64 @@
-MODULE accum_runoff_module
+MODULE accum_route_module
```
ekluzek, line 1: I didn't notice this before. But, the internal module name is different than the filename? That would be good to make consistent. This used to be required for the CESM build dependency generator, but it must not be true anymore.

route/build/src/base_route.f90
```diff
@@ -0,0 +1,60 @@
+MODULE base_route
+
+  ! Abstract class for routing method
```
ekluzek, line 3: I suggest a little more documentation here, to say that this is the base class for which specific routing methods will be extended from. It adds a subroutine that will need to be implemented by each specific class.

route/build/src/base_route.f90
```diff
@@ -0,0 +1,60 @@
...
+
+  implicit none
+
+  private
+  public:: base_route_rch
```
ekluzek, line 8: I'd add 1-line comments after each declaration

route/build/src/base_route.f90
```diff
@@ -0,0 +1,60 @@
...
+
+
+  ! --- routing method container
+  type, abstract :: base_route_rch
+    CONTAINS
+    procedure(sub_route_rch), deferred :: route
+  end type
+
+  ABSTRACT INTERFACE
+
```
ekluzek, line 24: I suggest adding some comments to talk a bit about this subroutine and what's required in it.

route/build/src/dfw_route.f90
```diff
@@ -1,135 +1,38 @@
...
+USE dataTypes,     ONLY: STRFLX          ! fluxes in each reach
+USE dataTypes,     ONLY: STRSTA          ! state in each reach
+USE dataTypes,     ONLY: RCHTOPO         ! Network topology
+USE dataTypes,     ONLY: RCHPRP          ! Reach parameter
+USE dataTypes,     ONLY: dwRCH           ! dw specific state data structure
+USE public_var,    ONLY: iulog           ! i/o logical unit number
+USE public_var,    ONLY: realMissing     ! missing value for real number
+USE public_var,    ONLY: integerMissing  ! missing value for integer number
+USE public_var,    ONLY: qmodOption      ! qmod option (use 1==direct insertion)
+USE globalData,    ONLY: idxDW
```
ekluzek, line 15: What is this? It doesn't have a one line description...

route/build/src/dfw_route.f90
```diff
@@ -1,135 +1,38 @@
...
+USE water_balance, ONLY: comp_reach_wb   ! compute water balance error
+USE base_route,    ONLY: base_route_rch
```
ekluzek, line 17: Probably good to define a one line comment that this is the base class.

route/build/src/accum_runoff.f90
```diff
@@ -1,147 +1,64 @@
...
  ! *********************************************************************
  ! subroutine: perform accumulate immediate upstream flow
  ! *********************************************************************
- SUBROUTINE accum_qupstream(iEns,       &    ! input: index of runoff ensemble to be processed
-                            segIndex,   &    ! input: index of reach to be processed
-                            ixDesire,   &    ! input: reachID to be checked by on-screen pringing
-                            NETOPO_in,  &    ! input: reach topology data structure
-                            RCHFLX_out, &    ! inout: reach flux data structure
-                            ierr, message)   ! output: error control
+ SUBROUTINE accum_inst_runoff(this,         &
```
ekluzek, line 31: Here and below I'd add something to say that the "this" is the object this subroutine operates on. Because this is a standard pattern it's not too important, but could be helpful for someone not familiar with object oriented programming. I tend to like more comments in general...

This could be done other places as well, and should just be consistent if it's done...

ekluzek (commented):
I think this is great. A nice improvement that actually decreases the amount of code and removes duplication, which is great. It also allows new methods to be added without touching the old ones which is nice for code management.

I have a few suggestions, mostly to add a few more comments. I also noticed something I didn't notice before which is a module that is different than it's filename. That might be something good to make consistent. But, it's not needed to do here, that could be seperate from this.

route/build/src/accum_runoff.f90
nmizukami: I will make module name consistent with file name.  Mostly both are the same, but some module name is different than the file name (I will correct later)

route/build/src/accum_runoff.f90
nmizukami: I can add something like this "accum_runoff_rchr" object to bound this procedure

route/build/src/dfw_route.f90
nmizukami: added the comment - "routing method index for diffusive wave"

route/build/src/dfw_route.f90
nmizukami: adding comment: base (abstract) reach routing method class

route/build/src/base_route.f90
nmizukami: adding one comment for this and below

route/build/src/base_route.f90
nmizukami: adding descriptions of the routine

route/build/src/accum_runoff.f90
nmizukami: I am wondering if I call a module name xxx_mod for xxx_mod.f90 or no "mod"  if I put "_mod", I will need to change most of f90 files (I guess that is fine, but wondering what would be best).

nmizukami: So for now, I will leave this as and change later after solid decision

route/build/src/base_route.f90
nmizukami: add this comments

PR #421, opened 2023-08-11
connect lake module to the other routing methods

nmizukami:
Enable KW, MC, and DW routing method with lake routine.

Connecting KWT requires more changes in `kwt_route.f90` (implementation is not finished for KWT)

Picture1

### DIFF
--- a/route/build/src/historyFile.f90
+++ b/route/build/src/historyFile.f90
@@ -591,6 +591,30 @@ MODULE historyFile
         if(ierr/=0)then; message=trim(message)//trim(cmessage); return; endif
       endif
 
+      if (meta_rflx(ixRFLX%KWvolume)%varFile) then
+        if (nRch_write>0) then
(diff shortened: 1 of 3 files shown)


nmizukami:
lake discharge and volume at Lake Geneva
Picture2

ekluzek:
I'd just like to understand the dependencies between namelist options for cesm-coupling. It sounds like before this, some of the route_opt options couldn't be used, as it assumed which one was used. Now, it looks like route_opt can be anything but 0. This is the kind of thing that should be checked for in the scripting level, as well as in the code when the control file is read in. I'd like to add a check in the buildnml to make sure if lake is turned on, that only the allowed route_opt options can be choosen. It's good to make sure any other dependencies are handled similarly. The script level and the code level (when the control file is read or inside module specific initialization) should ensure the options work together.

ekluzek (approved):
Looks good to me.

PR #423, opened 2023-08-13
Fix error in writing history_file variable for gauge-only history file in restart file 

nmizukami:
I got error in restart file writing (`history_file` variable) when gauge history file writing option is on.

I think `start` variable in `write_char0D` had a bug.

### DIFF
--- a/route/build/src/pio_utils.f90
+++ b/route/build/src/pio_utils.f90
@@ -692,19 +692,18 @@ CONTAINS
   ierr=0; message='write_char0D/'
 
   if ( len(string) > size(tmpString) )then
-      ierr = 1
-      message=trim(message)//'ERROR: length of string being written is larger than tmpString'
-      return
+    ierr = 1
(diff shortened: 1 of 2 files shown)


route/build/src/pio_utils.f90
```diff
@@ -692,19 +692,18 @@ SUBROUTINE write_char0D(pioFileDesc,     &
...
   if(ierr/=0)then; message=trim(message)//'ERROR: getting variable id'; return; endif
 
-
   do m = 1,len(string)
-     tmpString(m:m) = string(m:m)
+    tmpString(m:m) = string(m:m)
   end do
-  start(1) = iStart
-  start(2) = 1
+  start(1) = 1
```
ekluzek, line 705: I think you are right here. The first dimension is the number of characters and the second is the number of history files.

ekluzek (approved):
Yes, this looks right to me. I just checked to make sure.

PR #429, opened 2023-09-15
Enable to run at user specified routing time step rather than using coupling_frequency

nmizukami:
**This is for CESM coupled run.**

So far, mizuRoute runs at the same time step as coupling time step (== coupling frequency, now set to 1 days). IRF routing is ok to run at whatever time step.   `dt_qsim` In control file is reset to coupling frequency (`coupling_period`) in initializing phase.

With this PR,  `dt_qsim` In control file is kept and mizuRoute run at this time step. `dt_qsim` is smaller than `coupling_period`, mizuRoute run with the same runoff input at `nsub` (`coupling_period/dt_qsim`) time, called sub-cycling.  `dt_qsim` should be less than 86400 second.

One caveat. 
Now, after sub-cycling, river volume is computed at last sub-time step is exported to the coupler, but what should be exported would be average over the coupling period?

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -12,7 +12,7 @@ MODULE RtmMod
                           ice_runoff, do_rof, do_flood, &
                           river_depth_minimum, &
                           nsrContinue, nsrBranch, nsrStartup, nsrest, &
-                          cfile_name, coupling_period, &
+                          cfile_name, coupling_period, delt_coupling, nsub, &
                           caseid, brnch_retain_casename, inst_name, &
                           barrier_timers
   USE RunoffMod,    ONLY: rtmCTL, RunoffInit         ! rof related data objects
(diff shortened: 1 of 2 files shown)


route/build/cpl/RtmMod.F90
```diff
@@ -118,16 +119,30 @@ SUBROUTINE route_ini(rof_active,flood_active)
...
 
     !-------------------------------------------------------
-    ! 1. Initialize time
+    ! 1. Get ROF subcycling
     !-------------------------------------------------------
-    ! If routing time step dt [sec] and coupling time step coupling_period [day] is different, match dt to coupling_period.
-    if (dt/=coupling_period) then
+    delt_coupling = coupling_period*secprday   ! day -> sec
+
+    if (abs(mod(delt_coupling, dt)-0._r8)>verySmall) then
```
ekluzek, line 126: I'm glad you are doing this checking or otherwise I'd suggest it. How big is verySmall in this case it likely should be zero, as you really need it to be exact.

ekluzek (approved):
This looks good to me. I have one question that I ask. I think for time we really have no tolerance for not being exact. So might be something to think about...

route/build/cpl/RtmMod.F90
nmizukami: Hi Erik, verySmall=tiny(1.0_dp).   so smallest number resolved by double precision?

nmizukami: replace `verySmall` with `0.0_dp`

PR #431, opened 2023-09-29
Refactoring related to PIO decomposition initialization

### LINKED ISSUE #409: Only create PIO decompositions once...
A small performance improvement would be to create the PIO decompositions only once, and only free them when the model is shutting down. Right now we create them for each new file. I don't think this is that important to do, but would be modestly better.

### LINKED ISSUE #438: Defining pio system
Currently `pioSystem` (global variable in mizuRoute) is defined as a _type_ in globalData.F90 [here](https://github.com/ESCOMP/mizuRoute/blob/13520fecaaf74ad62ad160f763d97e231357b558/route/build/src/globalData.f90#L142. ).  

This shared variable is used for both standalone run and cesm-coupling run. 

For standalone, `pioSystem` variable is created as a _type_ [here](https://github.com/ESCOMP/mizuRoute/blob/0a1afbe97cbb5ed14d3fe00a2995158b1c36ca96/route/build/src/standalone/model_setup.f90#L96) and used throughout the run

For cesm-coupline, `pioSystem` is defined in the coupler as a _pointer_ and assigned [here](https://github.com/ESCOMP/mizuRoute/blob/0a1afbe97cbb5ed14d3fe00a2995158b1c36ca96/route/build/cpl/RtmMod.F90#L187C4-L187C52). 
since the function shr_pio_getiosys(inst_name) return a pointer, `pioSystem` should have been a pointer. So this creates inconsistency. 

If `pioSystem` in globalData is defined as a pointer, standalone run produce the error, but if `pioSystem` in globalData is defined as a pointer, both standalone and cesm-coupled run work ok. So currently it is defined as a type, but unsure if this will break in the future or this breaks with other compilers than the ones we test. 

This is also discussed in PR #431. 

We will consult with Jim

nmizukami:
Move pio decomposition initialization to the model initialization (beginning of the model execution). Idea is to define decomposition data once and share pio system (which decomposition data is attached to) for all the history files and restart file, and avoid defining decomposition data every time the output files are created.

This does not change any science and should re-produce the results from the codes prior to this PR.

resolves #409
resolves #438

### DIFF
--- a/route/build/src/globalData.f90
+++ b/route/build/src/globalData.f90
@@ -139,7 +139,7 @@ MODULE globalData
   integer(i4b),                    public :: pio_rearranger    = 2              ! 0=>PIO_rearr_none 1=> PIO_rearr_box 2=> PIO_rearr_subset
   integer(i4b),                    public :: pio_root          = 1              ! PIO root
   integer(i4b),                    public :: pio_stride        = 1              ! PIO stride - see PIO documentation for more information
-  type(iosystem_desc_t),           public :: pioSystem                          ! PIO I/O system data
+  type(iosystem_desc_t), pointer,  public :: pioSystem                          ! PIO I/O system data
   ! pio decomposition used for history file variables
   type(io_desc_t),                 public :: ioDesc_hru_float                   ! [hru] (float)
   type(io_desc_t),                 public :: ioDesc_rch_float                   ! [reach] (float)
--- a/route/build/src/globalData.f90
+++ b/route/build/src/globalData.f90
@@ -133,31 +133,32 @@ MODULE globalData
   integer(i4b),                    public :: nThreads                           ! number of threads
   logical(lgt),                    public :: masterproc                         ! root logical. root processor => true, other => false
   logical(lgt),                    public :: multiProcs                         ! MPI multi-processors logical. => number of processors>1 true, other => false
-  character(len=strLen),           public :: pio_netcdf_format = "64bit_offset" !
-  character(len=strLen),           public :: pio_typename      = "pnetcdf"      !
-  integer(i4b),                    public :: pio_numiotasks    = -99            !
+  character(len=strLen),           public :: pio_netcdf_format = "64bit_offset" ! netCDF format - use '64bit_offset' for PIO use or 'netCDF-4'
+  character(len=strLen),           public :: pio_typename      = "pnetcdf"      ! netcdf, pnetcdf, netcdf4c, or netcdf4p
+  integer(i4b),                    public :: pio_numiotasks    = -99            ! Number of iotasks (ntasks/stride) - see PIO documentation for more information
   integer(i4b),                    public :: pio_rearranger    = 2              ! 0=>PIO_rearr_none 1=> PIO_rearr_box 2=> PIO_rearr_subset
-  integer(i4b),                    public :: pio_root          = 1              !
-  integer(i4b),                    public :: pio_stride        = 1              !
-  type(iosystem_desc_t), target,   public :: pioSystem                          ! PIO I/O system data
+  integer(i4b),                    public :: pio_root          = 1              ! PIO root
+  integer(i4b),                    public :: pio_stride        = 1              ! PIO stride - see PIO documentation for more information
+  type(iosystem_desc_t),           public :: pioSystem                          ! PIO I/O system data
   ! pio decomposition used for history file variables
-  type(io_desc_t),                 public :: ioDesc_hru_float
-  type(io_desc_t),                 public :: ioDesc_rch_float
-  type(io_desc_t),                 public :: ioDesc_gauge_float
-   ! decomposition used for restart variables
-  type(io_desc_t),                 public :: ioDesc_rch_int
-  type(io_desc_t),                 public :: ioDesc_rch_double
-  type(io_desc_t),                 public :: ioDesc_hist_rch_double
-  type(io_desc_t),                 public :: ioDesc_hru_double
-  type(io_desc_t),                 public :: ioDesc_wave_int
-  type(io_desc_t),                 public :: ioDesc_wave_double
-  type(io_desc_t),                 public :: ioDesc_mesh_kw_double
-  type(io_desc_t),                 public :: ioDesc_mesh_mc_double
-  type(io_desc_t),                 public :: ioDesc_mesh_dw_double
-  type(io_desc_t),                 public :: ioDesc_irf_double
-  type(io_desc_t),                 public :: ioDesc_vol_double
-  type(io_desc_t),                 public :: ioDesc_irf_bas_double
-  integer(i4b),   allocatable,     public :: index_write_gage(:)                !
+  type(io_desc_t),                 public :: ioDesc_hru_float                   ! [hru] (float)
+  type(io_desc_t),                 public :: ioDesc_rch_float                   ! [reach] (float)
+  type(io_desc_t),                 public :: ioDesc_gauge_float                 ! [gauge points] (float)
+  ! decomposition used for restart variables
+  type(io_desc_t),                 public :: ioDesc_rch_int                     ! [reach x ensemble] (integer)
+  type(io_desc_t),                 public :: ioDesc_rch_double                  ! [reach x ensemble] (double precision)
+  type(io_desc_t),                 public :: ioDesc_hist_rch_double             ! [reach] (double precision)
+  type(io_desc_t),                 public :: ioDesc_hru_double                  ! [hru] (double precision)
+  type(io_desc_t),                 public :: ioDesc_wave_int                    ! [reach x max. number of waves x ensemble] (integer)
+  type(io_desc_t),                 public :: ioDesc_wave_double                 ! [reach x max. number of waves x ensemble] (double precision)
+  type(io_desc_t),                 public :: ioDesc_mesh_kw_double              ! [reach x Euler kinematic wave computational meshes x ensemble] (double precision)
+  type(io_desc_t),                 public :: ioDesc_mesh_mc_double              ! [reach x Muskingum-Cunge computational meshes x ensemble] (double precision)
+  type(io_desc_t),                 public :: ioDesc_mesh_dw_double              ! [reach x Duffusive wave computational meshes x ensemble] (double precision)
+  type(io_desc_t),                 public :: ioDesc_irf_double                  ! [reach x IRF future timer-steps x ensemble] (double precision)
+  type(io_desc_t),                 public :: ioDesc_vol_double                  ! [reach x time-bounds x ensemble] (double precision)
+  type(io_desc_t),                 public :: ioDesc_irf_bas_double              ! [reach x basin UH future time-steps x ensemble] (double precision)
+  integer(i4b),   allocatable,     public :: index_write_gage(:)                ! reach indices to gauge points w.r.t. distributed domains
+
   ! ---------- conversion factors -------------------------------------------------------------------
 
   real(dp),                        public :: time_conv                  ! time conversion factor -- used to convert to mm/s
--- a/route/build/src/globalData.f90
+++ b/route/build/src/globalData.f90
@@ -139,8 +139,25 @@ MODULE globalData
   integer(i4b),                    public :: pio_rearranger    = 2              ! 0=>PIO_rearr_none 1=> PIO_rearr_box 2=> PIO_rearr_subset
   integer(i4b),                    public :: pio_root          = 1              !
   integer(i4b),                    public :: pio_stride        = 1              !
-  type(iosystem_desc_t),           public :: pioSystem                          ! PIO I/O system data
-
+  type(iosystem_desc_t), target,   public :: pioSystem                          ! PIO I/O system data
+  ! pio decomposition used for history file variables
+  type(io_desc_t),                 public :: ioDesc_hru_float
+  type(io_desc_t),                 public :: ioDesc_rch_float
+  type(io_desc_t),                 public :: ioDesc_gauge_float
+   ! decomposition used for restart variables
+  type(io_desc_t),                 public :: ioDesc_rch_int
+  type(io_desc_t),                 public :: ioDesc_rch_double
+  type(io_desc_t),                 public :: ioDesc_hist_rch_double
+  type(io_desc_t),                 public :: ioDesc_hru_double
+  type(io_desc_t),                 public :: ioDesc_wave_int
+  type(io_desc_t),                 public :: ioDesc_wave_double
+  type(io_desc_t),                 public :: ioDesc_mesh_kw_double
+  type(io_desc_t),                 public :: ioDesc_mesh_mc_double
+  type(io_desc_t),                 public :: ioDesc_mesh_dw_double
+  type(io_desc_t),                 public :: ioDesc_irf_double
+  type(io_desc_t),                 public :: ioDesc_vol_double
+  type(io_desc_t),                 public :: ioDesc_irf_bas_double
+  integer(i4b),   allocatable,     public :: index_write_gage(:)                !
   ! ---------- conversion factors -------------------------------------------------------------------
 
   real(dp),                        public :: time_conv                  ! time conversion factor -- used to convert to mm/s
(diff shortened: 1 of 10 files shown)


route/build/src/globalData.f90
```diff
@@ -139,8 +139,25 @@ MODULE globalData
   integer(i4b),                    public :: pio_rearranger    = 2              ! 0=>PIO_rearr_none 1=> PIO_rearr_box 2=> PIO_rearr_subset
   integer(i4b),                    public :: pio_root          = 1              !
   integer(i4b),                    public :: pio_stride        = 1              !
-  type(iosystem_desc_t),           public :: pioSystem                          ! PIO I/O system data
-
+  type(iosystem_desc_t), target,   public :: pioSystem                          ! PIO I/O system data
+  ! pio decomposition used for history file variables
+  type(io_desc_t),                 public :: ioDesc_hru_float
```
ekluzek, line 144: Good to have comments about what these new variables are about...

route/build/src/globalData.f90
```diff
@@ -139,8 +139,25 @@ MODULE globalData
...
+  type(io_desc_t),                 public :: ioDesc_rch_float
+  type(io_desc_t),                 public :: ioDesc_gauge_float
+   ! decomposition used for restart variables
```
ekluzek, line 147: Something that would be good to have here is to be able to easily flip the storage type between float and double. This is something we want to do in the future. Restarts should always be double, but history you want to be able to easily choose between float and double. You might still have the explicit double and float variables. But, maybe for example you keep the ioDescRchHist variable, but have it assigned to the double or float specific variable at a higher level at initialization? Then to flip between the types you can do it at initialization and not have to change other parts of the code...

route/build/src/historyFile.f90
```diff
@@ -30,14 +27,14 @@ MODULE historyFile
...
-    logical(lgt)          :: gageOutput=.false.  ! flag to indicate this is at-gage-only output (== output subset of reaches)
-    type(iosystem_desc_t) :: pioSys              ! PIO system (this does not have to be initialize for each history file?)
-    type(file_desc_t)     :: pioFileDesc         ! PIO data identifying the file
-    type(io_desc_t)       :: ioDescRchFlux       ! PIO domain decomposition data for reach flux [nRch]
-    type(io_desc_t)       :: ioDescHruFlux       ! PIO domain decomposition data for hru runoff [nHRU]
+    character(FileStrLen)          :: fname               ! netCDF name
+    integer(i4b)                   :: iTime=0             ! time step in output netCDF
+    logical(lgt)                   :: fileStatus=.false.  ! flag to indicate history output netcdf is open
+    logical(lgt)                   :: gageOutput=.false.  ! flag to indicate this is at-gage-only output (== output subset of reaches)
+    type(iosystem_desc_t),pointer  :: pioSys              ! PIO system (this does not have to be initialize for each history file?)
```
ekluzek, line 34: I suggest removing pioSys from here, but use pioSystem from globaldata, and then make pioSystem in globalData a pointer, and change RtmMod so that you set pioSystem as a pointer like this...

    pioSystem         => shr_pio_getiosys(inst_name)

route/build/src/historyFile.f90
```diff
@@ -84,61 +78,36 @@ FUNCTION constructor(fname, pioSys, gageOutput) RESULT(instHistFile)
...
-      if (present(pioSys)) then
-          instHistFile%pioSys=pioSys
-      else
-        pio_numiotasks = nNodes/pio_stride
-        call pio_sys_init(pid, mpicom_route,          & ! input: MPI related parameters
-                          pio_stride, pio_numiotasks, & ! input: PIO related parameters
-                          pio_rearranger, pio_root,   & ! input: PIO related parameters
-                          instHistFile%pioSys)       ! output: PIO system descriptors
-      end if
+      instHistFile%pioSys => pioSys
```
ekluzek, line 81: Remove this just use pioSystem from globalData where needed...

route/build/src/historyFile.f90
```diff
@@ -327,45 +296,6 @@ SUBROUTINE closeNC(this)
       endif
     END SUBROUTINE closeNC
 
-    ! ---------------------------------
```
ekluzek, line 330: It's nice to not to HAVE to call of these as new files are created. However, you could keep this code around and call it only when mizuRoute is closing down.

route/build/src/historyFile.f90
```diff
@@ -515,105 +446,48 @@ SUBROUTINE write_flux_rch(this, hVars_local, index_write, ierr, message)
         if(ierr/=0)then; message=trim(message)//trim(cmessage)//' [array_temp]'; return; endif
       end if
 
-      ! write instataneous local runoff in each stream segment (m3/s)
```
ekluzek, line 518: Below looks like you were able to really simplify the code which is really great to see.

ekluzek (commented):
This is a really nice change to bring in, that makes the code a bit simpler, and should also improve I/O performance by not having to recreate these PIO objects with each new file. This becomes more important the more frequently files are created.

From looking at other CESM code it looks like making pioSystem a pointer in globalData is an important thing to do.

route/build/src/init_model_data.f90
```diff
@@ -28,9 +28,49 @@ MODULE init_model_data
...
+  integer(i4b), intent(in)    :: pid              ! proc id
+  integer(i4b), intent(in)    :: nNodes           ! number of procs
+  integer(i4b), intent(out)   :: ierr             ! error code
+  character(*), intent(out)   :: message          ! error message
+  ! local variables
+  character(len=strLen)       :: cmessage         ! error message from subroutine
+
+  ! pio system initialization
+  if (trim(runMode)=='standalone') then
+    pio_numiotasks = nNodes/pio_stride
```
nmizukami, line 61: Hi @ekluzek,   just below, I create pioSystem if it is standalone, so this should be actual object instead of pointer?   if it is cesm-coupling,  pioSystem is pointing to the one initialized in the coupler with `shr_pio_getiosys(inst_name)` in RtmMod.f90.

route/build/src/historyFile.f90
nmizukami: pioSys has been removed from this type definition and use pioSystem directly from globalData

route/build/src/historyFile.f90
nmizukami: this has been also removed since pioSys does not exist any more

route/build/src/historyFile.f90
nmizukami: So now that pioSys are pio decomposition descriptors) are not in the histFile type, I think it would be better to clean function outside this type??   I have clean function in pio_utits.f90

nmizukami:
Hi Erik, using `pioSystem` as pointer did not work. It compiled with both gnu and intel, but did not run for both. No obvious messages (I think it died when it tried to create the file, which uses `pioSystem`).   I am thinking it is the best to leave this as is (so leave `pioSystem` defined as actual type) and figure out what to do later since this works both stand-alone and cesm coupling now.  or any other idea?

ekluzek:
> Hi Erik (@ekluzek), using `pioSystem` as a pointer did not work. It compiled with both gnu and intel, but did not run for both. No obvious messages (I think it died when it tried to create the file, which uses `pioSystem`). I am thinking it is the best to leave this as is (so leave `pioSystem` defined as pio_system type, not pointer) and figure out what to do later since this works both stand-alone and cesm coupling now. or any other idea?

@nmizukami yeah if it works I guess we can leave it as is. I'm nervous about it because the CESM interface is returning a pointer and we aren't treating it that way. But, if it works, I guess it's OK. Something to check though, is to make sure it works for the compilers we want it to. And I'm also concerned that maybe the reason it works happens to be because of the internals of either PIO or the compilers we use -- so that if either changes -- it'll break. But, we can also handle that when it comes rather than now, especially since this is problematic. I'd also like to show this to Jim Edwards and see if he has thoughts on this. So we should schedule a meeting for that.

route/build/src/init_model_data.f90
ekluzek: This is in #431

route/build/src/globalData.f90
ekluzek: This will be worked on more later.

route/build/src/historyFile.f90
ekluzek: Ahh, OK, should be removed.

ekluzek (approved):
@nmizukami did a bunch of work on the things that I asked for which ended up being problematic. We are going to go with as is because it's working.

I will run some testing for CESM as is and we'll make a tag that will go into the next CESM version.

ekluzek:
@nmizukami I ran the tests on cheyenne and they were all fine running as expected. The tests on izumi had some tests pass, but the NAG tests showed problems...

ERS_D_Mmpi-serial.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_nag.mizuroute-default    RUN
SMS_D.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_nag.mizuroute-default        RUN
SMS_D_Mmpi-serial.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_nag.mizuroute-default    RUN

With DEBUG mode turned on in above, you'll see things like dangling pointers caught for example in that middle one. 

Since, the last one is the simplest this is what you see captured in the traceback for cesm.log:


``` text
Completion(send) value=1072724297 tag=1
Completion(send) value=977411098 tag=1
Completion(send) value=1199980800 tag=1
Completion(sRuntime Error: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/historyFile.f90, line 407: Reference to dangling pointer THIS%IODESCHRUFLUX
Target was RETURNed from procedure HISTORYFILE:SET_COMPDOF_RCH_HRU
Program terminated by fatal error
/fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/historyFile.f90, line 407: Error occurred in HISTORYFILE:WRITE_FLUX_HRU
/fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/historyFile.f90, line 372: Called by HISTORYFILE:WRITE_FLUX
/fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/write_simoutput_pio.f90, line 210: Called by WRITE_SIMOUTPUT_PIO:OUTPUT
/fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/cpl/RtmMod.F90, line 676: Called by RTMMOD:ROUTE_RUN
/fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/cpl/nuopc/rof_comp_nuopc.F90, line 792: Called by ROF_COMP_NUOPC:MODELADVANCE
/fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/cmeps/cime_config/../cesm/driver/esmApp.F90, line 141: Called by ESMAPP
end) value=1124738608 tag=1

```

nmizukami:
Thank you Erik.   

So I believe this is nmizukami:cesm-coupling_pio-decomp branch (make sure I am looking at the same code)?  

Dangling pointer means that a target is deallocated (or does not exist for some reason)?  I wonder if I actually need to clean (nullify the pointer) when a history netcdf is finished and closed before new history file (= before histFile object is created) like this:

```
    SUBROUTINE cleanup_rch_hru(this)
      implicit none
      class(histFile), intent(inout) :: this

      nullify(this%ioDescRchFlux)
      nullify(this%ioDescHruFlux)

    END SUBROUTINE cleanup_rch_hru
```

Other than that, I don't see apparent error in the code.

ekluzek:
Hi @nmizukami. I looked at this some more to see a little bit of what's happening. I was going to try your idea of nullifying pointers, but I realized that this isn't happening at the end -- it's fairly early on. But, it's also after other successful output happens.

Here's a code snippet where this is happening...

```
      call write_netcdf(this%pioFileDesc, 'time', [hVars_local%timeVar(1)], [this%iTime], [1], ierr, cmessage)

      call write_netcdf(this%pioFileDesc, 'time_bounds', hVars_local%timeVar, [1,this%iTime], [2,1], ierr, cmessage)

      if (.not.this%gageOutput) then
        if (meta_hflx(ixHFLX%basRunoff)%varFile) then
          call this%write_flux_hru(hVars_local, ierr, cmessage).    !!! <<<< dying in here
        endif
      end if

      call this%write_flux_rch(hVars_local, index_write, ierr, cmessage)

      call sync_file(this%pioFileDesc, ierr, cmessage)

    END SUBROUTINE write_flux
```

ekluzek:
Here's my casedir on izumi...

/scratch/cluster/erik/tests_mizu_c-cpln3_v211_ctsm51d114m/SMS_D_Mmpi-serial.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_nag.mizuroute-default.GC.mizu_c-cpln3_v211_ctsm51d114m_nag

nmizukami:
Ok, so nullify the pointer did not work. So need to think something else...   do you think something like [this](https://github.com/jacobwilliams/json-fortran/issues/307) is related to this error?  I was googling with "fortran dangling pointer" and ran into this.

route/build/src/init_model_data.f90
nmizukami: Allocating `pioSystem` before calling pio_sys_init to define it works.

nmizukami:
> > Hi Erik (@ekluzek), using `pioSystem` as a pointer did not work. It compiled with both gnu and intel, but did not run for both. No obvious messages (I think it died when it tried to create the file, which uses `pioSystem`). I am thinking it is the best to leave this as is (so leave `pioSystem` defined as pio_system type, not pointer) and figure out what to do later since this works both stand-alone and cesm coupling now. or any other idea?
> 
> @nmizukami yeah if it works I guess we can leave it as is. I'm nervous about it because the CESM interface is returning a pointer and we aren't treating it that way. But, if it works, I guess it's OK. Something to check though, is to make sure it works for the compilers we want it to. And I'm also concerned that maybe the reason it works happens to be because of the internals of either PIO or the compilers we use -- so that if either changes -- it'll break. But, we can also handle that when it comes rather than now, especially since this is problematic. I'd also like to show this to Jim Edwards and see if he has thoughts on this. So we should schedule a meeting for that.

Hi Erik,  I change `pioSystem` to a pointer in globalData.f90.  As Jim said, allocating `pioSystem` before calling `pio_sys_init` works for standalone.  cesm-coupling mode also works (I just tested one case - f09_f09_rMERIT with cheyenne intel.

nmizukami:
Hi Erik,  I updated handling of the pio decomposition data for history file output.  At least there are no pointers (except for pioSystem), so will need to test with nag compiler in izumi.   At cheyenne, it works as expected.

ekluzek:
@nmizukami I ran that one test on izumi, and it PASSes! So I'll rerun all the testing which hopefully will all work, and then we can merge this and make a tag!

nmizukami:
> @nmizukami I ran that one test on izumi, and it PASSes! So I'll rerun all the testing which hopefully will all work, and then we can merge this and make a tag!

Ah, great!  thank you for testing! Yes, we can merge and tag it.  

I also want to setup and run ctsm-mizuroute in izumi (something I want to talk about today).

ekluzek:
Tests on izumi all PASS, the Cheyenne tests are going:

ERI_Mmpi-serial.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_gnu.mizuroute-default
ERS_D_Mmpi-serial.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_nag.mizuroute-default
PET_Mmpi-serial_P1x25.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_intel.mizuroute-default
SMS_D.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_intel.mizuroute-default
SMS_D.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_nag.mizuroute-default
SMS_D_Mmpi-serial.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_gnu.mizuroute-default
SMS_D_Mmpi-serial.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_intel.mizuroute-default
SMS_D_Mmpi-serial.5x5_amazon_r05.I2000Clm50SpMizGs.izumi_nag.mizuroute-default

ekluzek:
OK, tests on Cheyenne are as expected. So this is ready to merge and create a tag.

PR #442, opened 2023-11-10
cheyenne build readme update

nmizukami:
Update the instruction in README_cheyenne_build 

Still need to figure what derecho minimum modules are required to build the model. 

Also updated testCase and its link


nmizukami:
Hi Erik,  we are moving to derecho soon, so wondering if this update is useful...  wondering how the derecho configurations should be included (just update this readme later?)

PR #448, opened 2023-12-28
enabling to use the other routing methods than IRF

nmizukami:
With this PR,  one of the other routing methods than IRF can be used for cesm-coupling mode

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -32,6 +32,7 @@ MODULE RtmMod
 
   implicit none
   logical, parameter :: verbose=.false.
+  integer, parameter :: iRoute=1        ! index of routing method chosen. cesm-coupled run allows only one option (5 options available), this is always 1.
 
   private
   public route_ini          ! Initialize mizuRoute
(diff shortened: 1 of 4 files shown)


route/settings/SAMPLE-coupled.control
```diff
@@ -76,7 +76,11 @@
 <MCroutedRunoff>        T                          ! kinematic wave routed discharge at reach [L3/T]
```
ekluzek, line 76: I suggest adding a comment header to this section that says something like

! --------------------
! Route discharge options
! Only one route option can be TRUE, others must be FALSE
! Routing options for types not chosen will be ignored.
!----------------------

route/build/cpl/RtmMod.F90
```diff
@@ -739,10 +738,10 @@ SUBROUTINE get_river_export_data(NETOPO_in, RCHFLX_in, &
         if (present(offset)) ix=ix+offset
         ! stream volume is split into HRUs based on HRU's areal weight for contributory area
         if (update_vol) then
-          rtmCTL%volr(ix) = RCHFLX_in(iens,iRch)%ROUTE(idxIRF)%REACH_VOL(1)*NETOPO_in(iRch)%HRUWGT(iHru)/rtmCTL%area(ix)
+          rtmCTL%volr(ix) = RCHFLX_in(iens,iRch)%ROUTE(1)%REACH_VOL(1)*NETOPO_in(iRch)%HRUWGT(iHru)/rtmCTL%area(ix)
```
ekluzek, line 741: As we talked about today, I'd like to have some explanation about why the index is hardcoded to "1". That could be in a comment for each line. But, better I think is to have an index called say something like "iRoute" that is fixed to 1, where you give the explanation that the route array is of size 1 for the cesm-coupled case. as only one routing type can be chosen. 

When I see a hardcoded "1" in an index I wonder why -- it's hardcoded to 1. So an explanation can be helpful.

ekluzek (approved):
As per our discussion earlier, I suggest a couple of changes for readability.

route/settings/SAMPLE-coupled.control
nmizukami: the note added.

route/build/cpl/RtmMod.F90
nmizukami: added `iRoute` parameter set to 1 and comments. then now using iRoute instead of hardcoded 1

Issue #396, opened 2023-06-01
floodplain implementation


nmizukami:
Now, mizuRoute uses a rectangular channel with unlimited channel depth. 

By implementing bankfull depth of the channel, allow the water to spill over the channel to the floodplain. This allow additional feedback of floodwater to Land model, as well as impact discharge simulation.

**Rough tasks**

_1 conceptualize floodplain._ 
simple one- sloped floodplain (1:S slope means height at 100 meter from channel edge is 100*S meter higher than channel top)
complex one- look at table (height vs area based on topography). See Yamazaki et al., 2011 WRR (floodplain inundation paper)

_2 need to implement new channel parameter_
 bankfull depth and also floodplain slope. MERIT-hydro may have this information, or somehow parameterize just like channel width.

_3 modify diffusive wave, kinematic wave (Euler one) and muskingum cunge routing routine_
Use bankfull storage, and if computed water volume exceeds this, flood happens and compute water height based on water spilled over the floodplain. Based on the water height adjusted by flood, compute celerity and diffusivity (for diffusive wave) and change flow rate.   As far as I see, IRF is not working with floodplain.  KWT may not also. 

4 hydraulic variables
final hydraulic variables are discharge, water volume, flood water volume, water height. 

**Reference:**

Yamazaki, D., Kanae, S., Kim, H., and Oki, T. (2011), A physically based description of floodplain inundation dynamics in a global river routing model, Water Resour. Res., 47, W04501, doi:[10.1029/2010WR009726](https://doi.org/10.1029/2010WR009726).

PR #449, opened 2024-01-03
Adding parameterized floodplain 

nmizukami:
**Purpose:** remove unlimited bankful depth so that water can go to flatter floodplain next to the channel.

**Assumption:** floodplain is the same for both sides. channel geometry is known shapes (rectangular, triangular, trapezoidal)

**Main code changes:**
- Add a routine to compute channel hydraulic properties (top of channel width, flow area, wetting perimeter, uniform flow, normal flow depth, celerity, and diffusivity) 
- Add a bankful depth, channel side slope, and flood plain slope to channel geometry

### DIFF
--- a/route/build/Makefile
+++ b/route/build/Makefile
@@ -162,6 +162,7 @@ DATATYPES = \
     allocation.f90
 # define utilities
 UTILS = \
+    hydraulic.f90 \
     mpi_utils.f90 \
     model_utils.f90 \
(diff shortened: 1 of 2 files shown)


nmizukami:
related to #396, but not resolved yet

nmizukami:
This PR does not affect any current model simulations and performance since the new module is not hooked up with any others).

PR #451, opened 2024-01-04
mizuRoute_control.py change required

nmizukami:
module called `six` is removed due to cime update.


nmizukami:
I just merged this because of the small change

Issue #359, opened 2023-03-31
PIO error when using gnu (> v10.1.0) and MPT


nmizukami:
When using gnu compiler with MPT, PIO sync fails (seemingly randomly) as segmentation fault (invalid memory reference).   Using intel compiler with MPT works fine.  Using gnu with openmpi works fine (seems to be).  I have been running into this problem for long time (for several years now). 

More specific configuration is:
gnu v12.1.0 
netcdf v 4.8.1 
pnetcdf v1.12.3
mpt v2.25
 
trace back looks like this.  14 through 25 are not displayed but it would be in C codes. 
piolib_mod.F90 Line 1372 is just PIOc_sync(file%fh) 

```
#13  0x2b9d2f8c8f66 in PMPI_File_write_at_all
	at /usr/src/packages/BUILD/mpt/lib/libmpi/src/romio/mpi-io/write_atall.c:61
#14  0xc53728 in ???
#15  0xc3ae8f in ???
#16  0xc38984 in ???
#17  0xc3a4f2 in ???
#18  0xc369ce in ???
#19  0xc37203 in ???
#20  0xb99763 in ???
#21  0x7b8fc1 in ???
#22  0x7b365e in ???
#23  0x7b917b in ???
#24  0x78559b in ???
#25  0x7077a9 in __piolib_mod_MOD_syncfile
	at /glade/u/home/mizukami/sandbox_mizuRoute/libraries/parallelio/src/flib/piolib_mod.F90:1372
#26  0x4193f2 in __pio_utils_MOD_sync_file
	at /glade/u/home/mizukami/sandbox_mizuRoute/route/build/../build/src/pio_utils.f90:391
#27  0x46dcc8 in __historyfile_MOD_write_flux
	at /glade/u/home/mizukami/sandbox_mizuRoute/route/build/../build/src/historyFile.f90:483
#28  0x58a35e in __write_simoutput_pio_MOD_output
	at /glade/u/home/mizukami/sandbox_mizuRoute/route/build/../build/src/write_simoutput_pio.f90:224
#29  0x7042d8 in route_runoff
	at /glade/u/home/mizukami/sandbox_mizuRoute/route/build/../build/src/standalone/route_runoff.f90:81
#30  0x7043f7 in main
	at /glade/u/home/mizukami/sandbox_mizuRoute/route/build/../build/src/standalone/route_runoff.f90:11
[...]

PR #453, opened 2024-01-24
PIO sync for potentially better robustness for gnu/cray-mpich and a few cleanups

nmizukami:
We found that cesm-coupling.n02_v2.1.2 produces the history output problem for the gnu (12.2.0) / mpich (8.1.27) compiled executable. pio_sync was added after writing the history file and this reduces the occurrence of this error (but not completely eliminating this problem).  Note: Jim mentioned when netcdf is closed, pio_sync happens so no explict call for pio_sync is needed. 

Note: This problem does occur only for the executable compiled with gnu/mpich. gnu/openmpi (5.0.0) and intel/mpich work fine. 

Further debugging for gnu/mpich is needed, but at this point, unclear about root cause of the problem. This might be related to #359 

Switching netcdf type from `pnetcdf` to `netcdf` clear this problem (for gnu (12.2.0) / mpich (8.1.27)). Therefore, expose the `pio_netcdf_type` to user control file. so user can specify netcdf type without re-compiling the code (`pio_netcdf_type` was hard-coded as `pnetcdf`.)

Fixed library (PIO) Makefile bug that cause the compiling error when you use only netCDF and gcc.

Other changes include removing unused variables or routines

### DIFF
--- a/route/build/Makefile
+++ b/route/build/Makefile
@@ -166,6 +166,7 @@ DATATYPES = \
     allocation.f90
 # define utilities
 UTILS = \
+    hydraulic.f90 \
     mpi_utils.f90 \
     model_utils.f90 \
     gamma_func.f90
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -32,6 +32,7 @@ MODULE RtmMod
 
   implicit none
   logical, parameter :: verbose=.false.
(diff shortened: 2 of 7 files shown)


nmizukami:
At least using standard netcdf for PIO works for gcc with mpich (not parallel-netcdf), and several compiling issues for recent compilers (both intel and gcc)

- [ ] issue on gcc and intel with mpich using standard parallel netcdf. 
- [ ] issue on cmake flag
- [ ] issue on recent gcc compile (Rank mismatch between actual arguments)

So merge this now.

PR #461, opened 2024-04-06
 makefile cleanups

nmizukami:
Removed unnecessary libraries - cime, mct, mpi-serieal for makefile and externals.cfs.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -1,17 +1,3 @@
-[cime]
-local_path = cime
-protocol = git
-repo_url = https://github.com/ESMCI/cime
-tag = cime6.0.156
-required = True
-
-[mct]
-tag = MCT_2.11.0
-protocol = git
-repo_url = https://github.com/MCSclimate/MCT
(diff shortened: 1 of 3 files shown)


nmizukami:
Hi Erik @ekluzek,  just wanted to make sure that these cleanup does not affect cesm coupling.

ekluzek (approved):
This is all great. We talked about this a few weeks ago and this is implementing those things we talked about.

And @nmizukami -- the answer is yes -- this won't affect CESM coupling at all. So from the CESM/CTSM perspective this is completely safe.

nmizukami:
Need a better error handling netcdf specification, but I will figure out how to do in Makefile. 

Now, either NCDF_PATH or ( NCDF_C_PATH AND NCDF_FORTRAN_PATH) needs to be provided, depending on how each system handle netcdf library (separate PATH for C and Fotran or not). if neither is provided, compilation fails, which happened before this PR.  

So right now this PR is merged and closed.

PR #464, opened 2024-04-19
Making intel-oneAPI compiler work 

nmizukami:
Convert do concurrent to regular do loop. 

No answer changes and this would be more stable.


ekluzek (approved):
This is a good idea to do anyway, since the "do concurrent" here, may interfere with the either MPI distributed memory parallelism or the OpenMP shared memory parallelism.

PR #467, opened 2024-05-17
Enabling reach routing at headwater HRUs

nmizukami:
Currently, reach routing in a **headwater** HRU is not performed because lateral flow (`BASIN_QR`) is always added at the bottom of reach. Note nothing is coming from the top of the headwater reach! 

This PR provides an option to add lateral flow at the top of reach in the headwater, so the headwater reach has volume of water as well as discharge from the reach. 

New control variable is `hw_drain_point` (integer)-- 1-> add top of reach, 2-> add bottom of reach.  Default is 2 as currently done.


nmizukami:
Changed MC routing volume check, resulting in slight change in discharge (nccmp detect this difference)

Issue #450, opened 2024-01-03
CESM moving from use of manage_externals to "git submodules" in cesm3_0_beta01


ekluzek:
CESM is considering moving from the use of manage_externals to using "git submodules". The CTSM PR that will do this for CTSM is here: 

https://github.com/ESCOMP/CTSM/pull/2443

Since, the mizuRoute externals are only used for standalone cases it doesn't need to happen the same time as other CESM components. But, should likely be done fairly soon afterwards as manage_externals is already NOT being supported at this point.

ekluzek:
This is happening in roughly a month. So we'll need a PR to do this in mizuRoute. We can check with Jim if he is willing to do it, or if we will need to do it. I think it's not too hard either way.

ekluzek:
Since, for mizuRoute we only have one external, and it's only needed for offline mizuRoute doing this does NOT need to happen immediately. So when we have a chance we'll update this for offline mizuRoute, but doing this isn't tied to any CESM timelines.

ekluzek:
cesm3_0_beta01 has been made where the move to git-fleximod has been done for all other CESM components. So we should now bring this in for mizuRoute.

- [ ] Add "git subtree" for git-fleximod as in other CESM components
- [ ] Update README_EXTERNALS file to mention git-fleximod
- [ ] Update Makefile that mentioned manage_externals
- [ ] Put above into a PR and merge to cesm-coupling
- [ ] Test that git-fleximod functions as we want
- [ ] Bring in #468 that removes manage_externals

PR #468, opened 2024-05-29
remove manage_externals

jedwards4b:
Remove manage externals from the mizuRoute repository

Part of #450


ekluzek:
@jedwards4b thanks so much for putting this PR together. We appreciate your doing it.

Note, that we weren't planning on making this a requirement for cesm3_0_beta01, as the only need for externals management for mizuRoute is for the standalone version. So we figure this can come in after cesm3_0_beta01 is done.

nmizukami:
Thanks Jim (@jedwards4b) for starting this, One thing I would like to check before merging this is:
do we have to have alternative to pull externals? Here PIO. I cannot remember what the new method is to replace manage external. git submodule??

jedwards4b:
@nmizukami git-fleximod is the replacement.  I can add it to this PR but won't do it now as apparently this update is to be deferred.

ekluzek:
We also need to bring in git-fleximod as a subtree as documented here:

https://github.com/ESMCI/git-fleximod/wiki#including-git-fleximod-in-other-repositories

ekluzek:
@nmizukami will bring this in as is as the second step to moving to git-fleximod. We will do another PR to bring git-fleximod in as a git subtree.

ekluzek (approved):
This can happen now that git-fleximod was merged into mizuRoute.

PR #470, opened 2024-06-13
Including floodplain

nmizukami:
This PR includes:

- hydraulic properties that used to compute celerity and diffusivity (for DW routing) use uniform flow equation (i.e., manning equation). 
- Adding options to include floodplain for each reach. This requires bankfull depth, floodplain slope (default is 1% slope extending from river bank). 

This is similar approach to CaMAFlood, or HyMAT type floodplain mechanism. This does not intend to do HEC-RAS type channel geometry definition (with many channel cross sections information) that kills run for high resolution river network like MERIT hydro over global, due to memory and much high intense computation.

floodplain is not implemented to IRF (no concept of channel geometry) and KWT (seemingly more complex) at least for now

Default is no floodplain, if you want to activate floodplain, need to set <floodplain> to T in control file.

Address #396


route/build/src/process_ntopo.f90
```diff
@@ -212,26 +213,40 @@ SUBROUTINE augment_ntopo(nHRU,             & ! input: number of HRUs
...
+   if (.not.meta_SEG(ixSEG%sideSlope)%varFile) &
+     structSEG(iSeg)%var(ixSEG%sideSlope)%dat(1) = 0._dp                                                 ! channel side slope h:v=slope:v [-]. now 0->rectangular channel
+   if (.not.meta_SEG(ixSEG%floodplainSlope)%varFile) &
+     structSEG(iSeg)%var(ixSEG%floodplainSlope)%dat(1) = floodplainSlope                                 ! floodplain slope
+ end do
+
+ ! replace depth with large values e.g., 10,000 [m] if floodplain mode is off
+ if (.not.floodplain) then
+   do iSeg=1,nSeg
+     structSEG(iSeg)%var(ixSEG%depth)%dat(1) = 10000._dp ! [meter]
```
ekluzek, line 234: I suggest making 10000._dp it a parameter called something like large_number just to denote that the value doesn't matter. Maybe use huge?

route/build/src/public_var.f90
```diff
@@ -12,8 +12,6 @@ MODULE public_var
 
   save
 
-  ! ---------- common constants ---------------------------------------------------------------------
```
ekluzek, line 15: Why is this removed?

route/settings/SAMPLE-coupled.control
```diff
@@ -45,12 +45,6 @@
 <varname_downSegId>     Tosegment                  ! name of variable holding the ID of the next downstream segment
 <varname_pfafCode>      PFAF                       ! name of variable holding the pfafstetter code
 ! ****************************************************************************************************************************
-! Define options to include/skip calculations
```
ekluzek, line 48: Remove settings that are just for the standard defaults.

route/build/src/ncio_utils.f90
```diff
@@ -198,6 +199,25 @@ subroutine get_nc_dim_len(fname,           &  ! input: filename
 
  end subroutine
 
+ ! *********************************************************************
+ ! subroutine: check if variable exists
+ ! *********************************************************************
+ function check_variable(ncid, vname)
```
ekluzek, line 205: I like having a subroutine for this kind of functionality.

ekluzek: Would it be useful to add calls to this in other places as well, so that you could check if variables exist.

ekluzek (approved):
I did a quick scan with @nmizukami and had a couple suggestions. But, looks good as far as I can see from a quick scan from an SE angle.

Naoki is going to make sure this is working OK and bring in as is. Later development for the science and tuning of it could happen in the future. But, getting the base version in would allow it to be used.

route/build/src/public_var.f90
nmizukami: I removed this because this is duplicated comment with line 20.

route/build/src/process_ntopo.f90
nmizukami: yes, changed this number to huge(1.0_dp)

nmizukami (commented):
a few things changed based on the comments from Erik (Thanks!)

PR #477, opened 2024-08-13
write lake parameters only for a lake id specified in control file in a log file

nmizukami:
**Log file writing for lake variables:** 

Now, write out the lake parameters for _all the lakes_ with H06 type, overwhelming a log file.

With this PR, write them only for lake id specified at <desireId > in a control file.


nmizukami:
This is minor update and just clean up the codes

PR #478, opened 2024-08-13
H06 default parameters renamed

nmizukami:
Make H06 parameter default names (five of them) consistent with the name used in meta data. 

Note that users feel free to use any names in the netCDF and set the names in control file.


nmizukami:
This is also minor clean up, so merged.

PR #485, opened 2024-10-04
Adding git-fleximod

### LINKED ISSUE #450: CESM moving from use of manage_externals to "git submodules" in cesm3_0_beta01

nmizukami:
Added and enabled git-fleximod functionality

- Created subtree of git-fleximod.
- Replaced Externals.cfg with .gitmodule
- Created a git-fleximod wrapper in bin directory
- Updated a document (READM_EXTERNALS.rst)

This PR may need to be merged together with #468 

Resolve #450


ekluzek (approved):
LGTM

PR #486, opened 2024-10-21
Adding datetime in rpointer file with backward compatibility

### LINKED ISSUE #484: Add timestamp to rpointer files, but make it backwards compatible with files without the timestamp
CESM is going to a system wide change where the restart rpointer text files will have the simulation timestamp in the filename. However, the model will still be backwards compatible to use the older filenames without it. But, new rpointer files created by the model will have the timestamp in the filename.

PR's for CTSM and MOSART for this work are here:

https://github.com/ESCOMP/CTSM/pull/2757
https://github.com/ESCOMP/MOSART/pull/105

nmizukami:
Adding string of datetime (`yyyy-mm-dd-sssss` format) to the end of `rpointer.rof` for only cesm-coupling mode. This is the backward compatibility. standalone does not use this for now.

Resolve #484

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -811,29 +811,41 @@ CONTAINS
     USE RtmFileUtils, ONLY: relavu, getavu, opnfil
     USE RtmVar,       ONLY: inst_suffix
     USE public_var,   ONLY: rpntfil
+    USE globalData,   ONLY: simDatetime
 
     implicit none
     ! Arguments:
     character(len=*), intent(out) :: pnamer ! full path of restart file
     ! Local variables:
-    integer :: i                  ! indices
-    integer :: nio                ! restart unit
-    integer :: status             ! substring check status
-    character(len=256) :: locfn   ! Restart pointer file name
+    integer :: i                    ! indices
+    integer :: nio                  ! restart unit
+    integer :: sec_in_day           ! time in second
+    character(len=17) :: timestamp  ! datetime string in yyyy-mm-dd-sssss
+    integer :: status               ! substring check status
+    character(len=256) :: locfn     ! Restart pointer file name
     !--------------------------------------------------------
 
     if (masterproc) then
       write(iulog,*) 'Reading restart pointer file....'
     endif
 
-    nio = getavu()
-    locfn = './'// trim(rpntfil)//trim(inst_suffix)
+    nio = getavu() ! get available unit number
+
+    ! construct rpoint file name with datetime - simDatetime has three datetime at previous(0), current(1) and next(2) time stamp
+    sec_in_day = simDatetime(1)%hour()*60*60+simDatetime(1)%minute()*60+nint(simDatetime(1)%sec())
+    write(timestamp,'(".",I4.4,"-",I2.2,"-",I2.2,"-",I5.5)') &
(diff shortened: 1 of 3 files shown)


route/build/src/io_rpointfile.f90
```diff
@@ -18,20 +19,33 @@ MODULE io_rpointfile
...
+    character(*),   intent(out)          :: message       ! error message
+    type(datetime), intent(in), optional :: curDatetime   ! current datetime
+    integer                              :: sec_in_day    ! time in second
+    character(len=17)                    :: timestamp     ! datetime string in yyyy-mm-dd-sssss
+    character(len=strLen)                :: rpntfil_path  ! rpointer file path
 
     ierr=0; message='io_rpfile/'
 
+    if (present(curDatetime)) then
+      sec_in_day = curDatetime%hour()*60*60+curDatetime%minute()*60+nint(curDatetime%sec())
```
ekluzek, line 37: Change 60*60 to nint(secphour) and 60 to nint(secpmin) in public_vars

route/build/cpl/RtmMod.F90
```diff
@@ -811,29 +811,41 @@ SUBROUTINE restFile_read_pfile( pnamer )
...
 
     if (masterproc) then
       write(iulog,*) 'Reading restart pointer file....'
     endif
 
-    nio = getavu()
-    locfn = './'// trim(rpntfil)//trim(inst_suffix)
+    nio = getavu() ! get available unit number
+
+    ! construct rpoint file name with datetime - simDatetime has three datetime at previous(0), current(1) and next(2) time stamp
```
ekluzek, line 834: change rpoint to rpointer

route/build/cpl/RtmMod.F90
```diff
@@ -811,29 +811,41 @@ SUBROUTINE restFile_read_pfile( pnamer )
...
+    character(len=256) :: locfn     ! Restart pointer file name
     !--------------------------------------------------------
 
     if (masterproc) then
       write(iulog,*) 'Reading restart pointer file....'
     endif
 
-    nio = getavu()
-    locfn = './'// trim(rpntfil)//trim(inst_suffix)
+    nio = getavu() ! get available unit number
```
ekluzek, line 832: Note, that new FORTRAN constructs allow you to NOT need getavu() anymore. There's a way to open a file with an available unit.

This would be a good modernization effort to do. But, not completely required.

ekluzek: I think the only thing in RtmFileUtils that is maybe useful is opnfil, but the use of getavu and relavu could be removed and handled within opnfil.

ekluzek (approved):
Had a couple comments/suggestions. And a larger thing that I'll make an issue about.

route/build/cpl/RtmMod.F90
nmizukami: Thanks for suggestions on ascii file handling. I will clean this up with another PR. This RtmFileUtils and ascii_utils could be merged and several routines (getavu and relavu, getSpareUnit) can be removed.

route/build/src/io_rpointfile.f90
nmizukami: fixed.

PR #487, opened 2024-10-22
sync mpi task before processing tmp file

nmizukami:
if using lots of tasks, writing tmp file from each task and reading tmp file are racing, and some tasks read empty tmp file, and get error. Adding mpi barrier seems to fix this problem.

### DIFF
--- a/route/build/src/standalone/model_setup.f90
+++ b/route/build/src/standalone/model_setup.f90
@@ -181,6 +181,8 @@ CONTAINS
                        inputFileInfo,    & ! output: input file information
                        ierr, message)      ! output: error control
 
+  USE globalData,          ONLY: masterproc, &
+                                 mpicom_route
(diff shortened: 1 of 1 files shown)


nmizukami:
Hi @ekluzek , 
yes, I can explore the other solutions. Yes, this is very beginning of the code, so I did not notice any slow down by using barrier here. 

The problem (I suspected) was while some tasks are creating tmp file (the same information in tmp file), some tasks are reading the same tmp file. So some tasks may get empty tmp file (if one task is creating and other task is reading at the same time). So wait till all the tasks are finishing creating tmp file before reading it. 

I was thinking of using one task to create a tmp file, but having all the tasks read it to populate date structure in all the tasks rather than having one task read and broadcast them, because the data type is a little bit too complex for each task to put them in the data type. but this way still need a barrier so make sure a tmp file exist when tasks read a tmp file.

PR #491, opened 2024-11-05
Clean up

### LINKED ISSUE #490: Remove RtmFileUtils.F90
RtmFileUtils is outdated now that FORTRAN has better ways built into manage file units and such. There isn't anything in there that's important to keep. And there already is a ascii_utils.F90 that likely covers anything that might be useful.

getSpareUnit in ascii_utils could also be removed as said above.

Something like this can be done...

``` fortran
      open( newunit=nu_nml, file=trim(NLFilename), status='old', iostat=nml_error )
.
.
.
      close(nu_nml)
```

nmizukami:
1. lib/mpi-serial directory is removed as this is not necessary 

2. update on file unit number
Use `newunit` for open file so there is no need to look for open unit number manually.

3. (standalone only). Now all the tasks create a tmp file listing all the runoff netcdf paths based on wildcard, and then read the temp file to populate a array with runoff netcdf paths.  This may cause racing condition (creating vs reading). To avoid racing condition, need mpi barrier. Instead of using mpi barrier, have a main task only do this sequence of the processes up to populate the array, and then broadcast the array to the other tasks. 

resolve #490


route/build/src/mpi_utils.f90
```diff
@@ -334,6 +335,58 @@ SUBROUTINE shr_mpi_bcastLogical(allocArray,    & ! inout: array to be broadcaste
...
+    character(strLen),              intent(out)   :: message       ! error message
+    ! local variable
+    character(len=:), allocatable                 :: flat_str
+    integer(i4b)                                  :: flat_size
+    integer(i4b)                                  :: idx
+    integer(i4b)                                  :: num_elements
+
+    ierr=0; message='shr_mpi_bcastChar/'
+
+    if (masterproc) then
```
ekluzek, line 356: Suggest following how this is done in csm_share.

https://github.com/ESCOMP/CESM_share/blob/main/src/shr_mpi_mod.F90

PR #493, opened 2024-11-12
Fixing overflow error coming from default large bankfull depth

### LINKED ISSUE #492: river storage when floodplain is off - likely cause overflow
Check bunkful river storage when floodplain is off. 

Since bankful depth is huge, so computing maximum storage (which is a function of depth), is likely cause overflow. So the storage may be also set to huge. When floodplain is off, maximum storage is not really used.
 
https://github.com/ESCOMP/mizuRoute/blob/e8a4f2b98236d4177540ae6fa01b91a20b546d6c/route/build/src/process_ntopo.f90#L239

nmizukami:
Unlimited bankfull depth defined with huge function causes overflow error when computing hydraulic geometries in [hydraulic.f90](https://github.com/ESCOMP/mizuRoute/blob/cesm-coupling/route/build/src/hydraulic.f90). 

Use 100km height as a default huge depth.

Fix #492


nmizukami:
Also sneaking one minor change.

Remove print statement to tell which file is opened in pio. There are a few checks - a file exist or not, and then open the file or not.

PR #494, opened 2024-11-13
remove many mpi_barrier

### LINKED ISSUE #488: mpi_barrier clean up
`mpi_process.f90` contains many `shr_mpi_barrier`calls. Some of them may not be needed and can be cleaned up.

nmizukami:
Remove mpi barrier statement. Definitely no need barrier before especially before blocking calls (mpi_scatter, mpi_gather mpi_send, mpi_receive). 

Resolve #488

### DIFF
--- a/route/build/src/mpi_process.f90
+++ b/route/build/src/mpi_process.f90
@@ -329,8 +329,6 @@ CONTAINS
 
   endif  ! ( masterproc )
 
-  call shr_mpi_barrier(comm, cmessage)
-
   ! sends the number of reaches/hrus per proc to all processors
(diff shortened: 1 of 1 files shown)


ekluzek (approved):
This all looks good to me. Barriers are rarely needed. There are implicit blocking points already going on because of MPI activity like broadcasts, or gather/scatter where it has to wait for another task to do some specific thing.

So removing them is always, good unless there is a reason for them to be there. Which you will usually find when running.

@nmizukami looked at this together and didn't go into depth with each call, but the ones we did made sense for removal. He didn't find a difference in performance, nor did he find the model failing because of a lack of a barrier.

So good to go from my perspective.

Issue #476, opened 2024-08-07
Create new global HDMA-lake river input data including reservoir operation


nmizukami:
I got the data from Inne and need to examine and put her data into current HDMA-lake river network (`/glade/campaign/cgd/tss/people/mizukami/mizuRoute_data/cesm_mizuRoute/ancillary_data/Network_topology_HDMA_HydroLake_v3_cdf5.nc`)

and then test with CTSM

nmizukami:
it is `/glade/campaign/cgd/tss/people/mizukami/mizuRoute_data/cesm_mizuRoute/ancillary_data/Network_topology_HDMA_HydroLake_v4_cdf5.nc`

ekluzek:
Thing left to do is to move it to the standard place.

Also we'll need to have a way to control whether reservoir is on or off.

PR #496, opened 2024-11-23
Add capability to change lake type and bug fix

nmizukami:
Added control variable `<lakeRegulate>` with default T. 
`<lakeRegulate>` = `T`  Use `LakeModelType` defined in the input network data. 
`<lakeRegulate>` = `F`  Turn all the reservoir lakes into natural lake (using Doll to compute discharge).

If all the lakes are natural lakes (`LakeModelType`=1), no effect.
This does not affect endorheic lakes.

Also some public variables are not needed so removed

Also, found a bug in  subroutine `flow_depth` in hydraulic.f90. Flow depth computation needs to be performed only if input Q is greater than zero. Otherwise, flow depth is zero. Since `coef1` computation use division by input Q, so if Q is zero, error due to "divided by zero" happens when trying to compute the equations.

related to #476

### DIFF
--- a/route/build/src/process_ntopo.f90
+++ b/route/build/src/process_ntopo.f90
@@ -380,6 +380,7 @@ END SUBROUTINE augment_ntopo
   USE public_var,    ONLY: min_slope          ! minimum slope
   USE public_var,    ONLY: dt                 ! simulation time step [sec]
   USE public_var,    ONLY: is_lake_sim        ! lake simulation option
+  USE public_var,    ONLY: lakeRegulate       ! lake type option: T-> lake type defined at lake indiviually, F-> all natural (use Doll)
   ! external subroutines
   USE process_param, ONLY: basinUH            ! construct basin unit hydrograph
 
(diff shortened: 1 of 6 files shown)


route/build/src/process_ntopo.f90
```diff
@@ -396,6 +397,7 @@ SUBROUTINE put_data_struct(nSeg, structSEG, structNTOPO, &
   character(len=strLen)                          :: cmessage         ! error message of downwind routine
   integer(i4b)                                   :: nUps             ! number of upstream segments for a segment
   integer(i4b)                                   :: iSeg,iUps        ! loop indices
+  integer(i4b), parameter                        :: doll=1
```
ekluzek, line 400: Lake model type. If natural lake, it's doll.

This could use a comment to explain.

route/build/src/public_var.f90
```diff
@@ -101,9 +101,7 @@ MODULE public_var
   logical(lgt)         ,public    :: floodplain           = .false.         ! logical if flood water is computed or not (floodplain is added)
   integer(i4b)         ,public    :: hw_drain_point       = 2               ! how to add inst. runoff in reach for headwater HRUs. 1->top of reach, 2->bottom of reach (default)
   logical(lgt)         ,public    :: is_lake_sim          = .false.         ! logical if lakes are activated in simulation
-  logical(lgt)         ,public    :: lake_model_D03       = .false.         ! logical if Doll 2003 model is used, specify as 1 in lake_model_type in network topology
```
ekluzek, line 104: Nice simplification here by removing these logical variables.

ekluzek (approved):
Just one suggestion to add comments about the doll parameter.

But, nice simplification in removing the unneeded logical variables.

PR #498, opened 2024-12-02
use mid-point of time step for time stamp of history file for cesm-coupling run 

### LINKED ISSUE #413: Have time-stamps for "averaged" output history data be middle of interval, but "instantaneous" be end
This is important to comply with the latest convention decided in the CESM co-chairs group. time-stamps for history data that is averaged should be the middle of the interval, and time_bounds should be reported as the start and end of the interval. For instantaneous data (so history data output as "daily" with a daily time-step) the time-stamp should be the end of the interval, and time_bounds NOT included in output.

This was decided in co-chairs on 7/24/2023, and we want it to be put in place for all CESM components by end of Summer.

nmizukami:
CESM will require the component uses the mid-point of the output time step if average is used for aggregation.

For now, all the history file uses average for output variable aggregation in the history file. 

This PR enforces time stamp in the history file is the mid-point of the time stamp if `cesm-coupling` is on

resolve #413

### DIFF
--- a/route/build/src/public_var.f90
+++ b/route/build/src/public_var.f90
@@ -14,6 +14,8 @@ MODULE public_var
 
   ! physical constants
   real(dp),    parameter,public   :: pi=3.14159265359_dp    ! pi
+  real(dp),    parameter,public   :: Cw=4190.0_dp           ! heat capacity of water [J/kg/K]
+  real(dp),    parameter,public   :: RoW=0.99975_dp         ! density of water [kg/m³] at 10 C-degree
 
   ! some common constant variables (not likely to change value)
   real(dp),    parameter,public   :: secprmin=60._dp        ! number of seconds in a minute
--- a/route/build/src/write_simoutput_pio.f90
+++ b/route/build/src/write_simoutput_pio.f90
@@ -203,6 +203,10 @@ CONTAINS
        index_write_all = arth(1,1,nRch_local)
      end if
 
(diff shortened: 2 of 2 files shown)


route/build/src/write_simoutput_pio.f90
```diff
@@ -203,6 +203,10 @@ SUBROUTINE output(ierr, message)
        index_write_all = arth(1,1,nRch_local)
      end if
 
+     if (trim(runMode)=='cesm-coupling') then
```
ekluzek, line 206: Add a comment to just say that we are setting the timestamp of the average history output as the midpoint of the average.

route/build/src/public_var.f90
```diff
@@ -14,6 +14,8 @@ MODULE public_var
 
   ! physical constants
   real(dp),    parameter,public   :: pi=3.14159265359_dp    ! pi
+  real(dp),    parameter,public   :: Cw=4190.0_dp           ! heat capacity of water [J/kg/K]
```
ekluzek, line 17: Check that these agree with the current values set in the shr_const_mod.F90 code under the share directory.

ekluzek (approved):

route/build/src/write_simoutput_pio.f90
nmizukami: added comment: "For cesm coupling mode, the timestamp of the average history output is the midpoint of time bounds"

PR #506, opened 2025-02-03
Staging for direct insertion implementation

nmizukami:
This PR not meant to change anything existing config/run, but just adding new stuff without being called from any modules.

1. Adding three new f90 files. 
- gage meta object
- observation object (gague at reach or water extraction injection at reach)
- direct-insertion routine

2. A few minor changes in existing F90
- character(string?) array reading
- adding control variables related to direct insertion

These are to be used in the same data assimilation capability implemented in main branch

### DIFF
--- a/route/build/Makefile
+++ b/route/build/Makefile
@@ -176,6 +176,8 @@ DATATYPES = \
     var_lookup.f90 \
     base_route.f90 \
     csv_data.f90 \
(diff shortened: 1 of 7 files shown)


nmizukami:
This capability is needed soon or later (or now), so I will merge this and there is subsequent PR after this to enable direct insertion at gauge location.

PR #511, opened 2025-03-13
Added a new parameter, HYPE_Qsim_mode

ShervanGharari:
Introduced a new parameter, `HYPE_Qsim_mode`, in the HYPE reservoir formulation, allowing users to define reservoir outflow based either on the maximum of the primary and emergency spillway outputs or their sum.

This update improves the model's flexibility in simulating hydrograph peaks by enabling the inclusion of the emergency spillway in the reservoir outflow, even when its discharge is lower than the primary outflow. Previously, the outflow was determined solely by the maximum of the two spillways.


ShervanGharari:
Tested and resulted in identical simulation (IRF and HYPE lake formulation). Ready to be merged.

PR #522, opened 2025-03-18
Ensuring Lake Simulation Compatibility with Routing Options Beyond IRF (Cleaning QFUTURE_IRF and UH Allocation)

### LINKED ISSUE #517: Lake Simulation Fails with Routing Options Other Than IRF
This is based on my earlier discussion with @nmizukami and some testing on my end.

Lake simulation fails for routing options other than **IRF** (e.g., **DW**). The error occurs because, when IRF is not simulated, `NETOPO_in(segIndex)%UH` is not allocated in `process_ntopo.f90`, which causes the simulation to stop at [this line](https://github.com/ESCOMP/mizuRoute/blob/cesm-coupling/route/build/src/lake_route.f90#L473).

Upon some testing, it seems these [lines](https://github.com/ESCOMP/mizuRoute/blob/cesm-coupling/route/build/src/lake_route.f90#L472-L477) are redundant and can be fully removed or commented out in `lake_route.f90`. The lake does not need `QFUTURE_IRF` as memory since it is not used, and this memory is per segment, so it does not affect other segments to the best of my knowledge. Commenting out these lines of code in my tests resulted in identical streamflow simulations for a given case.

However, if **UH** or **QFUTURE_IRF** need to be allocated for other purposes, such as creating a restart file, I suggest the following:

- The **UH** can be set to an array of `[1, 0, 0]`, for example, if the segment is flagged as a lake in `process_ntopo.f90`.
- The **QFUTURE_IRF** can be allocated earlier before entering the lake subroutine. In my opinion, the best approach would be to have a routing-specific function just before calling the [lake subroutine](https://github.com/ESCOMP/mizuRoute/blob/cesm-coupling/route/build/src/main_route.f90#L301) to adjust any routing-specific tasks that do not affect the lake simulation itself, including allocating **QFUTURE_IRF** or any other related routing decisions. Therefore, the routing-specific code can be removed from `lake_route.f90`.

Let me know your thoughts. I can update the source code accordingly to resolve this issue.

PS: If relevant, I am using the GNU compiler.

ShervanGharari:
resolve #517


route/build/src/init_model_data.f90
```diff
@@ -418,6 +419,9 @@ SUBROUTINE init_state_data(pid, nNodes, comm, ierr, message)
       end if
       if (onRoute(impulseResponseFunc)) then
         do ix = 1,nRch_root
+          ntdh = size(NETOPO_trib(ix)%UH)
+          allocate(RCHFLX_trib(iens,ix)%ROUTE(idxIRF)%QFUTURE_IRF(ntdh), stat=ierr, errmsg=cmessage)
```
nmizukami, line 423: can combine allocation and initialization like this.
`allocate(RCHFLX_trib(iens,ix)%QFUTURE_IRF(ntdh), source=0._dp, stat=ierr, errmsg=cmessage)`

ShervanGharari:
> Hi Shervan, No, QFUTURE_IRF is not under ROUTE. Should be `RCHFLX_trib(iens,ix)%QFUTURE_IRF(1:ntdh)`

Yes, makes sense, thank you, and I'll update.

ShervanGharari:
Compiles, and results in identical simulation with original branch using IRF and HYPE reservoir formulation.
It now works with other routing options (tested with DW).
Does not output DW volume although the `islake` is on. Should I open an issue for that?

nmizukami:
> Does not output DW volume although the islake is on. Should I open an issue for that?
Did you add this in control file? Default should be F now
`<DWvolume>       T      ! history file output variable option`

ShervanGharari:
> > Does not output DW volume although the islake is on. Should I open an issue for that?
> > Did you add this in control file? Default should be F now
> > `<DWvolume>       T      ! history file output variable option`

Yes, I have already tried this and it works. But I meant to be automated when islake is True. Similar to `IRF`. Sorry for not clarifying this. As you decide.

nmizukami:
currently default meta_rflx(ixRFLX%IRFvolume)%varFile is True in popMetadat.f90. That is why you don't have to do anything to output IRFvolume if IRF is on, regardless whether lake is on or off. You could turn on volume output for all methods, but not always want to have volume in output.

ShervanGharari:
> currently default meta_rflx(ixRFLX%IRFvolume)%varFile is True in popMetadat.f90. That is why you don't have to do anything to output IRFvolume if IRF is on, regardless whether lake is on or off. You could turn on volume output for all methods, but not always want to have volume in output.

Understood, I'll create a new issue for that. I believe we can adopt a unified approach for IRFvolume, DWvolume, and other components to ensure consistency across routing options.

nmizukami:
I can merge this one. Is there anything to change for this particular PR (@ShervanGharari )?

ShervanGharari:
> I can merge this one. Is there anything to change for this particular PR (@ShervanGharari )?

Hi @nmizukami Thank you very much, I have nothing to add more to this PR. If it compiles and results in identical simulations on your side please merge. Thank you again for your support on this PR.

PR #523, opened 2025-03-19
Update testlist for derecho and fix a few problems

ekluzek:


nmizukami:
Hi @ekluzek, thanks for finishing derecho test case and fixing the rpointer file I/O (which looks good to me). I will look over the rest, and merge it (unless I have any question)

ekluzek:
Let me run the full test list first. And there is still one more task to do on this PR.

Then I'll remove the WIP label and have you merge it. I also want to have us make a tag with it as soon as we run the test list.

route/build/cpl/RtmFileUtils.F90
```diff
@@ -149,7 +149,7 @@ subroutine opnfil (locfn, iun, form)
      ! !ARGUMENTS:
      implicit none
      character(len=*), intent(in):: locfn  !file name
-     integer, intent(in):: iun             !fortran unit number
+     integer            :: iun             !fortran unit number
```
nmizukami, line 152: is this supposed to be input intent??  or inout??

nmizukami (commented):
Hi Erik, I started looking at this.  I don't have anything I noticed, except one inline comment on `opnfil` subroutine

route/build/cpl/RtmFileUtils.F90
ekluzek: It seems to need to be inout, so I removed the intent. But, I could be explicit to say

` integer, intent(inout):: iun`

if you prefer. That might be better because it means the intent was thought about. 

And actually, maybe what it should be is: intent(out), since that's the way it's really working. So I'll try that...

Issue #515, opened 2025-03-14
Backward Compatibility with `<fname_qsim>     file_names.txt` in control file


ShervanGharari:
Hi,
It looks like the handling of input file names has changed slightly in `cesm-coupling`:
```
<fname_qsim>       nc_file_*.nc  ! for multiple files using wildcard
```
or
```
<fname_qsim>       nc_file.nc    ! for a single file
```
Currently, the source code generates a list of NetCDF files and writes it to a temporary text file, using that as the file list instead of directly reading from `file_names.txt`. As a result, the following no longer works:
```
<fname_qsim>       file_names.txt  ! specifying a file list
```
I understand this change, given that cesm-coupling is now the main development branch and should align with the existing setups. However, it might be useful to support both approaches. If `file_names.txt` is provided instead of an NC file or a wildcard pattern, it could be copied directly to the temporary file rather than generating a new one [here](https://github.com/ESCOMP/mizuRoute/blob/cesm-coupling/route/build/src/standalone/model_setup.f90#L228).
If this approach is acceptable, I can incorporate the change into the code and submit a pull request.
Hope I am not missing something here and thank you very much in advance.

nmizukami:
any reasons for wanting this text file input??  do you have some netcdf names which cannot be specified using wild card?

This change was suggested by Andy (@andywood), but I am fine with allowing both text file input.

andywood:
@ShervanGharari -- having the required file list (a la summa) just introduced one more input file to generate or keep track of, which is unnecessary in many applications (i.e., if you have just a single runoff file, as you might in an ensemble short range forecast application.  I thought we could just let the code do the work rather than force it on the user, eg by letting it either expand a wildcard name to get its file list, or if it gets a 'txt' extension to the filename it would read that as the pre-made file list, or if it gets an .nc or .ncdf extension, it assumes that is the single runoff input file.   

Btw, that some of the NCAR development work is supporting non-CESM coupling applications, such as forecasting, so multiple usage options will likely need to be supported if the two very divergent current primary branches/forks are brought back together.

ShervanGharari:
Hi @andywood,
Thank you very much for the suggestion. After discussing with @nmizukami this morning, we concluded that we can easily accommodate both options in the code.
[...]

PR #524, opened 2025-03-20
Enhancing `<fname_qsim>` Support: Accepting a single NetCDF file (`file_name.nc`), multiple files via wildcards (`file_name*.nc`), or a text file listing filename(s) (`file_name(s).txt`)

ShervanGharari:
Enhancing `<fname_qsim>` Support: Accepting a single NetCDF file (`file_name.nc`), multiple files via wildcards (`file_name*.nc`), or a text file listing filename(s) (`file_name(s).txt`).
Resolving #515

### DIFF
--- a/route/build/src/ncio_utils.f90
+++ b/route/build/src/ncio_utils.f90
@@ -21,6 +21,7 @@ public::end_def
 public::write_nc
 public::open_nc
 public::close_nc
+public::is_netcdf_file
 
 INTERFACE get_nc
   module procedure get_scalar
--- a/route/build/src/ncio_utils.f90
+++ b/route/build/src/ncio_utils.f90
@@ -958,17 +958,17 @@ CONTAINS
  ! *********************************************************************
  ! Public subroutine: check if the file is NetCDF
  ! *********************************************************************
- SUBROUTINE is_netcdf_file(fname, is_nc, ierr, message)
+ function is_netcdf_file(fname, ierr, message) result(is_nc)
 
    implicit none
    ! input
    character(*), intent(in)  :: fname    ! Filename
    ! output
-   logical(lgt), intent(out) :: is_nc    ! Flag: True if NetCDF file
    integer(i4b), intent(out) :: ierr     ! Error code
    character(*), intent(out) :: message  ! Error message
    ! local
    integer(i4b)              :: ncid     ! NetCDF file ID
+   logical(lgt)              :: is_nc    ! Flag: True if NetCDF file
 
    ! initialize outputs
    is_nc = .false.
@@ -987,6 +987,6 @@ CONTAINS
      message = trim(message) // '[' // trim(nf90_strerror(ierr)) // '; file=' // trim(fname) // ']'
    end if
 
- END SUBROUTINE is_netcdf_file
+ end function is_netcdf_file
 
 END MODULE ncio_utils
--- a/route/build/src/ncio_utils.f90
+++ b/route/build/src/ncio_utils.f90
@@ -970,20 +970,20 @@ CONTAINS
    integer(i4b)              :: ncid     ! NetCDF file ID
 
    ! initialize outputs
-   is_nc = .FALSE.
+   is_nc = .false.
    ierr  = 0
    message = 'is_netcdf_file: '
 
    ! Try to open the file in read-only mode
-   ierr = nf90_open(TRIM(fname), NF90_NOWRITE, ncid)
+   ierr = nf90_open(trim(fname), NF90_NOWRITE, ncid)
 
    if (ierr == NF90_NOERR) then
      ! Successfully opened, so it's a NetCDF file
-     is_nc = .TRUE.
-     call nf90_close(ncid) ! close the file
+     is_nc = .true.
+     ierr = nf90_close(ncid) ! close the NetCDF file
    else
      ! Not a valid NetCDF file
-     message = TRIM(message) // '[' // TRIM(nf90_strerror(ierr)) // '; file=' // TRIM(fname) // ']'
+     message = trim(message) // '[' // trim(nf90_strerror(ierr)) // '; file=' // trim(fname) // ']'
    end if
 
  END SUBROUTINE is_netcdf_file
--- a/route/build/src/ncio_utils.f90
+++ b/route/build/src/ncio_utils.f90
@@ -902,7 +902,7 @@ CONTAINS
 
 
  ! *********************************************************************
- ! Public subroutine: close netcdf
+ ! Public subroutine: open netcdf
  ! *********************************************************************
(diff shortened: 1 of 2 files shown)


ShervanGharari:
@nmizukami the source code compiles; I still need to test it with all three options (file_name.txt, file_name.nc and file_name?/*.nc). Can you review and give me your initial feedback if you agree with the changes? Thank you in advance!

ShervanGharari:
Note: for some reasons that I do not fully understand, `ls Runof?.nc > tmp` is not working inside the Fortran code as it literally puts the string `Runof?.nc` to `tmp` file (while it works fine on command line outside of mizuRoute). I tried to add `bash -c` but it did not help. Better to test on your side as well, it can be dependent on the HPC config I am using.
The simulations are successful with `file_nam*.nc`, `file_name.nc`, and `file_name.txt`.

nmizukami:
> h

usually wildcard ? represents a single character.

ShervanGharari:
> usually wildcard ? represents a single character.

You are right! In the control file, I was missing a character. I was using `Runo?.nc` instead of `Runof?.nc`. Now works! Thank you and ready to be merged from my side.

ShervanGharari:
Two thoughts regarding the temporary file handling:

1- It might be better to place the temporary file in the folder where the compiled source is located rather than in the input folder. This would prevent issues in cases where the input folder is read-only, making it impossible to create the temporary file inside it.

2- I have simplified the process to always create or copy to a temporary file; however, this step can be skipped in certain cases (except for wildcard scenarios). For example, Andy suggested exploring ways to use the file without writing it to disk. One possible approach is to bypass temporary file creation when the input file is in NetCDF (NC) format.

I can easily implement both adjustments if you think they would improve the process. Let me know your thoughts.

route/build/src/ncio_utils.f90
```diff
@@ -953,4 +954,39 @@ SUBROUTINE close_nc(ncid, ierr, message)
 
  END SUBROUTINE close_nc
 
+
+ ! *********************************************************************
+ ! Public subroutine: check if the file is NetCDF
+ ! *********************************************************************
+ SUBROUTINE is_netcdf_file(fname, is_nc, ierr, message)
```
nmizukami, line 961: this could be a function instead of subroutine.

nmizukami (commented):
i think it would be easier (for me) to follow if you first check `file_name` is ascii or not in the beginning and then if it is, just open/read to get netcdf names from the ascii. if not, create tmp file including netcdf names. if you have a specific netcdf in input, you could directly add this to `datalines`.

nmizukami:
I would like to get some second opinion from @ekluzek (hopefully this is a simple enough for you to follow. This is only a few F90 file changes)

Erik, so this PR is stand-alone only, but tries to make an adjustment to allow users to specify either 1) runoff netcdf name(s) with wildcards (e.g., `runoff_195[0-9].nc`,` runoff*.nc` or even a single netcdf like `runoff.nc`) or 2) a name of ascii file listing all the runoff netcdfs in order to provide runoff input names. Before PR, only 1) was allowed.

ShervanGharari:
> i think it would be easier (for me) to follow if you first check `file_name` is ascii or not in the beginning and then if it is, just open/read to get netcdf names from the ascii. if not, create tmp file including netcdf names. if you have a specific netcdf in input, you could directly add this to `datalines`.

@nmizukami Thank you, Naoki. I have implemented the following steps as per your suggestion:
First, check if the file is a NetCDF file. If it is, allocate a single line to `dataLines` and assign the NetCDF filename to it. Also as @andywood suggested (no file is written on disk).
If the file is not a NetCDF file, verify its existence (assuming it is an ASCII file). If it does not exist, handle wildcards and create a `tmp` file in the same directory as the mizuRoute executable (instead of the input folder, which may be read-only).
If a file list is provided in the input folder, read it into `dataLines` and add the input folder name to each variable/lin inside `dataLines`
Finally, the code verifies that each line in `dataLines` corresponds to an existing file otherwise, it returns an error.
The source code compiles and I have checked it under various conditions.

nmizukami:
Hi @ShervanGharari , that logic sounds good to me and the actual changes looks good to me, but I will try to have Erik to take a quick look (maybe by Wed) before merge.

route/build/src/ncio_utils.f90
nmizukami: @ekluzek also said this would be better for a function. just return one thing given input, but you could still give output for error message using intent(out). 

This is something like below (not tested)

``` fortran
logical(lgt) function is_netcdf_file(fname, ierr, message)   
   implicit none
   ! input
   character(*), intent(in)  :: fname    ! Filename
   ! output
   integer(i4b), intent(out) :: ierr     ! Error code
   character(*), intent(out) :: message  ! Error message
   ! local
   integer(i4b)              :: ncid     ! NetCDF file ID

   ierr  = 0; message = 'is_netcdf_file: '

   ! Try to open the file in read-only mode
   ierr = nf90_open(trim(fname), NF90_NOWRITE, ncid)

   if (ierr == NF90_NOERR) then
     ! Successfully opened, so it's a NetCDF file
     is_netcdf_file = .true.
     ierr = nf90_close(ncid) ! close the NetCDF file
   else
     ! Not a valid NetCDF file
     is_netcdf_file = .false.
     message = trim(message) // '[' // trim(nf90_strerror(ierr)) // '; file=' // trim(fname) // ']'
   end if

 end function is_netcdf_file
```

nmizukami (commented):
I talked to Erik briefly, and only one comment. If you have energy to change this netcdf utility function, I could wait. Otherwise could merge and change this later. The changes in model_setup.F90 looks good and easy to follow now. Make sure you tested this and works for all the cases - single netcdf name, netcdf names with wildcards and text file?

route/build/src/standalone/model_setup.f90
```diff
@@ -212,29 +213,69 @@ SUBROUTINE inFile_pop(dir_name,         & ! input: name of the directory of the
...
+    ! check if the provided input file is netcdf
+    call is_netcdf_file (infilename, is_nc, ierr_dummy, message_dummy)
+    if (is_nc) then
+      ! allocate the dataLines
+      allocate(dataLines(1))
+      ! pass the infilename as netcdf directly
+      dataLines(1) = infilename
+    else
+      ! check if the file list exists, assumed to be ascii
+      INQUIRE(FILE=infilename, EXIST=infile_exists)
```
nmizukami, line 248: could change INQUIRE to lowercase

ShervanGharari:
@nmizukami Hi Naoki, thank you and @ekluzek. I've made the changes you mentioned. The source code successfully compiles and finishes the simulation in all three cases (`file_list.txt`, `nc_name.nc`, `nc_*/????.nc`). Please also check on your side before merging `.nc`. Thank you again for your input.

Issue #534, opened 2025-04-22
Make st_archive archive river model+DART output files


kdraeder:
Doing data assimilation with CESM+DART generates many new types of output files,
which st-archive should archive, assuming they are named according to CESM conventions.
Many of the new files will be handled by the esp component code, but others are
more closely associated  with the geophysical components and should be archived
with their files. We've chosen new file names in such a way that this can be done
with very small changes to
   $component/cime_config/config_archive.xml
   cime/CIME/XML/archive_base.py
A second small change is to make st-archive handle compressed files (*.gz).
This has no effect on model calculations or existing output.

We don't yet have an interface for doing data assimilation with mizuroute (or MOSART or RTM),
but we hope to, and I'm updating the config_archive files for most of CESM's other components,
so it seems efficient to do the river model(s) at the same time.
I've read that RTM is only used in paleo contexts, and MOSART is deprecated,
so I'd like advice about whether to make similar changes to the archiving in either of those.

I developed and tested the changes in a recent (2025-4-22) cesm3.0_alphabranch.
I've committed them to a feature branch based on master and pushed it to my fork of mizuRoute.
I can include more details here, or in a PR when it's time for that.

I don't have permission to add Assignees, Labels, etc., so I'm hoping that someone will fill those in,
or give me permission to.

ekluzek:
Cool, thanks @kdraeder. Can you make a PR to mizuRoute with the proposed change? I can see that you'll be adding just a couple new lines to config_archive.xml, but I'm not completely clear what it will look like. If you can create a PR, we can merge it in pretty readily. So your doing a PR would really help here.

And by the way, the same is true for CTSM, and technically if we see one example, we can figure it out for other submodules. But, it also helps if we have someone who creates the PR's for these things.

ekluzek:
Oh, and @kdraeder we are in the process of upgrading the license to Apache 2.0, so we'll need you to agree that having your changes under that new license is fine. See #500 for more details on this. We probably will bring this in before the license update, so we need to know about this for sure beforehand. If we bring this in after the change -- we still need you to realize that it's under the terms of the Apache 2.0 license. If you are like me -- you might not care about all this. But, we have to ask...

PR #535, opened 2025-04-23
Modify config_archive.xml in mizuroute to archive DART output files

kdraeder:
This would fix issue #534 
Testing is described in CIME PR https://github.com/ESMCI/cime/pull/4788.
Or I can import the details here.


cime_config/config_archive.xml
```diff
@@ -2,6 +2,7 @@
   <comp_archive_spec compname="mizuroute" compclass="rof">
     <rest_file_extension>r</rest_file_extension>
     <hist_file_extension>h.*\.nc$</hist_file_extension>
+    <hist_file_extension>e</hist_file_extension>
```
ekluzek, line 5: @kdraeder awesome, thanks for the PR here!

I think you should also add a test for this. In the test section below.

I think what that looks like is adding a line like this to the `<test_file_names>` section.

``` xml
<tfile disposition="move">casename.mizuroute.e.1976-01-01.nc</tfile>
```

ekluzek (approved):
We should add the test, but this is obviously very straightforward and makes sense!

cime_config/config_archive.xml
kdraeder: > @kdraeder awesome, thanks for the PR here!
> 
> I think you should also add a test for this. In the test section below.
> 
> I think what that looks like is adding a line like this to the `<test_file_names>` section.
> 
> ```
> <tfile disposition="move">casename.mizuroute.e.1976-01-01.nc</tfile>
> ```

I agree that there should be a test, but it's not clear to me that it's the right time.
The status of the river models now and in the near future wasn't clear to me, 
so the test I did for this series of PRs used MOSART.  So I didn't have mizuroute files to test.
Should I open a separate issue+PR for MOSART?

I could try to develop a test to generate the **standard** CESM mizuroute files.

As I noted in the issue, there's no interface from any river model to DART,
so I/we would also need to create some mizuroute **+DART** dummy files with the right names (.e.), 
and see whether st_archive handles them.  I have a script that can be adapted to make those.

nmizukami: Hi Kevin (@kdraeder), thank your for working on this! I am excited to see DA is used for river. mizuRoute is coupled only CTSM now (I compset?). if this is considered standard CESM mizuroute files, I think we have (or will have soon) the files @ekluzek? and wondering which mizuRoute grid should be used (there are two - vaguely speaking coarse and fine resolution, but they are unstructured grid). Also, note that mizuRoute is similar to wrf-hydro for river model (I think wrf-hydro is connected to DART?)

Yes, it is not too hard, I think issue+PR for MOSART would be good @ekluzek?

kdraeder: Hi Naoki,
Moha Gharamti (NCAR:CISL:DAReS) has developed DA for wrf-hydro, 
so when we find a motivated developer it should be reasonably straightforward to adapt 
that work to CESM:mizuRoute.  

This archiving issue doesn't depend on the mizuRoute grid, so any files with correct names 
will be useful.  I'd like to make sure that st_archive handles all of the files which a user
might want to save from mizuRoute; restarts, history, ...?

I don't know about the structure and expected lifespan of MOSART, 
so it's hard to say whether it's worth the effort to make it consistent with DART's file creation.

kdraeder: I heard from Jim Edwards today that MOSART will be the default river model in CESM3,
so I'll open an issue there to make similar changes.

ekluzek: Yes, MOSART is the default river model for CESM3, with RTM as the option for Paleo, and mizuRoute as an experimental setup for "I compset" cases. In the next CESM release we want to move to mizuRoute, and use it for both "I compset" and coupled cases as well as Paleo. And then remove RTM, and have MOSART in a backup position. There are some interested in continuing development in MOSART, so it might stick around for some. But, in order to remove the maintenance needs for supporting three river models we want to make mizuRoute the one that we put effort and support into.

ekluzek: @kdraeder the other thing is that I've put some effort into understanding how the st_archive tests work. And you should add the above test. The tests in that section are pure tests that just make sure the functionality works. So we should bring in it, as having it in place will make sure it's setup correctly as well as making sure it continues to work in the future.

kdraeder:
I've added the .e. line in the test section, following @ekluzek suggestion.

I'd like to test the config_archive.xml file, but I'm wondering how easy it is to set up a whole case
with active mizuRoute.  For my purposes I could make fake mizuroute files with the right names,
if I know what the names are.  If this seems like a reasonable plan, and someone can point me 
to a $rundir on /glade which has a representative set of output files, that would help.

cime_config/config_archive.xml
nmizukami: This is ready to be merged

PR #539, opened 2025-05-06
Remove ensemble dimension

### LINKED ISSUE #531: suggestion: ensemble dimension to be removed
Removing `ens` dimension in various data structures such as `RCHFLX`.  

Reasons:

- The ensemble run option has never been implemented; there is no ensemble runoff input implemented. So it is not usable now. Implementing this takes major changes. More complex code structure which make it harder to update in MPI.
- Less usage (No demand from a user) for ensemble run for last 10yrs.
- There is a better way to handle ensemble run using outside script (learn from CESM)

nmizukami:
Ensemble dimension has never been used (and never be implemented). 

Also, ensemble simulation (Note that ensemble based on multiple runoff forcing) can be performed with scripts that control mizuRoute executions with different forcing in parallel). This is easier implementation of parallel runs for ensembles. 

Fixes #531


ekluzek (approved):
By nature this is hard to review by eye. The changes make sense though and seem to be what needs to happen.

I suggest checking for uses of the words involved here with:

``` shell
git grep -i iens
git grep -i ensemble
git grep -i ndim_ens
git grep -i '%ens'
```
which should show if there are remaining changes to make

PR #549, opened 2025-06-26
Take 5: Documentation update

nmizukami:
Added local build capability. See docs/README.md 

ESCOMP/mizuRoute now granted permission for readthedoc webhook, so When anyone PR, it test building here. To test this, reference.rst was updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,4 @@ manage_externals.log
 
 # data folder
 /Hanasaki_test/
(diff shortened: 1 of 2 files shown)


nmizukami:
just merging this since content is ok and for some reason, readthedocs check here checked my PR test to my personal repo (PR 27 in my personal repo), which is wrong thing to test. it is supposed to check this PR.

PR #562, opened 2025-08-18
Add write_restart_at_endofrun functionality to rtm same as in clm

### LINKED ISSUE #557: mizuRoute needs handling to write restarts at end of run
The problem is described in MOSART in this issue, and the answer is going to be pretty identical in mizuRoute as in MOSART:

https://github.com/ESCOMP/MOSART/issues/123

nmizukami:
resolves #557

### DIFF
--- a/route/build/cpl/nuopc/rof_comp_nuopc.F90
+++ b/route/build/cpl/nuopc/rof_comp_nuopc.F90
@@ -758,6 +758,22 @@ contains
     call t_stopf ('lc_mizuRoute_import')
 
     !--------------------------------
+    ! Determine if time to stop
+    !--------------------------------
+
+    call ESMF_ClockGetAlarm(clock, alarmname='alarm_stop', alarm=alarm, rc=rc)
+    if (ChkErr(rc,__LINE__,u_FILE_u)) return
+
+    if (ESMF_AlarmIsRinging(alarm, rc=rc)) then
+       if (ChkErr(rc,__LINE__,u_FILE_u)) return
+       nlend = .true.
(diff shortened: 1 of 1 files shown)


ekluzek:
In CAM, what we did was to run a SMS, test, where we set the namelist variable:

```
write_restart_at_endofrun = .true.
```

in the user_nl_cpl

and then make sure it wrote out rpointer and restart files at the end. The SMS tests are setup with REST_OPTION='never' so without the above no restarts would have been written.

route/build/cpl/nuopc/rof_comp_nuopc.F90
```diff
@@ -750,16 +761,19 @@ subroutine ModelAdvance(gcomp, rc)
...
-       rstwr = .false.
+      call ESMF_ClockGetAlarm(clock, alarmname='alarm_restart', alarm=alarm, rc=rc)
+      if (ChkErr(rc,__LINE__,u_FILE_u)) return
+
+      if (ESMF_AlarmIsRinging(alarm, rc=rc)) then
+         if (ChkErr(rc,__LINE__,u_FILE_u)) return
+         rstwr = .true.
+         call ESMF_AlarmRingerOff( alarm, rc=rc )
+         if (ChkErr(rc,__LINE__,u_FILE_u)) return
+      endif
```
ekluzek, line 776: The other thing to change is to move the code to determine if it's time to stop to before the restart alarm.

ekluzek (approved):
One required change. But, hitting approve so you don't have to get back to me.

nmizukami:
looks like this is fixed. so merged

PR #564, opened 2025-09-14
write gauge id as character array in gauge only history file

nmizukami:
Now, when you write the results in gauge site only history file, the history file includes only reach ID. 

This PR includes gauge ID in the history file using character type in netcdf (not string which is available in netcdf4).


nmizukami:
just merged this.

PR #587, opened 2025-11-13
Update and fix regarding cesm-coupling part

nmizukami:
Fix a unit of discharge exported from ROF. It should be m/s, not m3/s. Also export discharge only at outlet elements.  Followed [MOSART](https://github.com/ESCOMP/MOSART/blob/0faa624a59b5ce677c40ed64f1ca65bb4997db8d/src/cpl/nuopc/rof_import_export.F90#L362)

Sample control file was updated


ekluzek (approved):
This fixes a problem with the budgets, because the units were wrong.

PR #592, opened 2025-12-01
Providing default names for input netcdfs

nmizukami:
Provide default names for:

- river network data (stand-alone & coupling)
- runoff or other input netcdf (stand-alone)
- mapping netcdf (stand-alone)

Update testCase in zonodo:   testCase_cameo_v3

### DIFF
--- a/route/build/src/ncio_utils.f90
+++ b/route/build/src/ncio_utils.f90
@@ -14,7 +14,7 @@ public::get_var_attr
 public::check_attr
 public::check_variable
 public::put_global_attr
-public::get_variable_rank
+public::get_variable_ndims
 public::def_nc
 public::def_dim
 public::def_var
@@ -203,14 +203,14 @@ CONTAINS
  end subroutine
 
  ! *********************************************************************
- ! subroutine: check if variable exists
+ ! subroutine: check number of dimension (rank) of a variable
  ! *********************************************************************
- function get_variable_rank(fname, vname)
+ function get_variable_ndims(fname, vname)
    implicit none
    ! input
    character(*), intent(in)        :: fname        ! filename
    character(*), intent(in)        :: vname        ! variable name
-   integer(i4b)                    :: get_variable_rank
+   integer(i4b)                    :: get_variable_ndims
    ! local
    integer(i4b)                    :: ierr         ! error code
    integer(i4b)                    :: ncid         ! NetCDF file ID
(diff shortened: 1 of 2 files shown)


route/build/src/ncio_utils.f90
```diff
@@ -14,6 +14,7 @@ MODULE ncio_utils
 public::check_attr
 public::check_variable
 public::put_global_attr
+public::get_variable_rank
```
ekluzek: When I see rank in a method name I figure it has something to do with MPI. But this is really about getting the PIO/netCDF ID. So I would recommend changing the name a bit.

ekluzek: It looks like both ndims and rank is used. So you could change to get_variable_ndims or add a comment to the subroutine.

route/build/src/tracer.f90
```diff
@@ -204,126 +204,126 @@ SUBROUTINE comp_mass_flux(idxRoute,       & ! input: routing method index
  END SUBROUTINE comp_mass_flux
 
 
- ! *********************************************************************
```
ekluzek: Maybe just delete this as git will have it in history.

route/build/src/public_var.f90
```diff
@@ -118,23 +118,23 @@ MODULE public_var
...
-  character(len=strLen),public    :: dname_nhru           = ''              ! dimension name of hru in river network data
+  character(len=strLen),public    :: fname_ntopNew        = charMissing     ! new filename containing stream network topology information
+  character(len=strLen),public    :: dname_sseg           = 'seg'           ! dimension name of segment in river network data
+  character(len=strLen),public    :: dname_nhru           = 'hru'           ! dimension name of hru in river network data
   ! RUNOFF, EVAPORATION AND PRECIPITATION FILE
   character(len=strLen),public    :: fname_qsim           = charMissing     ! runoff netCDF name
-  character(len=strLen),public    :: vname_qsim           = charMissing     ! variable name for runoff
-  character(len=strLen),public    :: vname_evapo          = charMissing     ! variable name for actual evapoartion
-  character(len=strLen),public    :: vname_precip         = charMissing     ! variable name for precipitation
+  character(len=strLen),public    :: vname_qsim           = 'runoff'        ! variable name for runoff
```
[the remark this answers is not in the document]
ekluzek: Naoki is thinking to unify around a standard set of names, so that it won't be needed most of the time.

route/build/src/ncio_utils.f90
nmizukami: changed "rank" to "ndims".

PR #596, opened 2025-12-24
Corrections related to CESM/CTSM coupling

### LINKED ISSUE #594: mizuRoute NUOPC cap only allows running with a daily or more coupling period
Currently the NUOPC cap gets the coupling period in days as an integer, which means trying to run subdaily ends up with 0 which is disallowed.

@nmizukami and I looked at this together, and the call to ESMF_TimeIntervalGet just needs to change from

d=coupling_period

to

s=coupling_period

And we may use s_r8 instead to return a double precision value. Which would work in the code easier.

nmizukami:
A few fixes related to cesm coupling:

1. Previously the coupling frequency was allowed to be more than 1 day because ``coupling_period`` was integer and unit of the coupling frequency was set to day in nuopc. Here, the type of ``coupling_period`` was changed to real and unit of frequency is set to sec. This allows us to use sub-daily coupling frequency and also make the code slightly cleaner (removed some lines)

2. Also there was bug (wrong loop index variable) when checking the outlet reach, which is fixed here. This fix seems to enable mizuRoute to run in fully CESM coupled mode without crash. 

Fix #594

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -12,7 +12,7 @@ MODULE RtmMod
                           ice_runoff, do_rof, do_flood, &
                           river_depth_minimum, &
                           nsrContinue, nsrBranch, nsrStartup, nsrest, &
-                          cfile_name, coupling_period, delt_coupling, nsub, &
+                          cfile_name, coupling_period, nsub, &
                           caseid, brnch_retain_casename, inst_name, &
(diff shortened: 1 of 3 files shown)


ekluzek (approved):
We worked on this together, and it makes sense as to what needs to be done.

Mostly fine detail and nothing needed to change or add documentation or notes on.

PR #601, opened 2026-01-27
Adding an optional input to enforce minimum lateral runoff in main_route subroutine

nmizukami:
**Background:**
Currently minimum lateral runoff volume into a river reach is set to 1.e-15 m3/s with `runoffMin` ([public_var.f90](https://github.com/ESCOMP/mizuRoute/blob/8d487fcfbcbeaf7b5d212a4f29a1de18f55c5f2f/route/build/src/public_var.f90#L31)), after depth is converted to volume. This threshold, even though it is very small, causes water imbalance (i.e., artificially creating runoff in the river model through the simulation periods) when running within fully coupled ESM if HRUs have zero lateral inflow (which happens at location with ocean mask in land model but land mask in river model). Therefore, it is appropriate to enforce minimum runoff input to zero. 

**What is in this PR:**

1. Change the default `runoffMin` to be zero, but a user can still change it to whatever minimum in the control file (`<runoffMin>`).
2. Add an optional input argument in [`mpi_route`](https://github.com/ESCOMP/mizuRoute/blob/8d487fcfbcbeaf7b5d212a4f29a1de18f55c5f2f/route/build/src/mpi_process.f90#L1142) subroutine, then  pass it to [`main_route`](https://github.com/ESCOMP/mizuRoute/blob/8d487fcfbcbeaf7b5d212a4f29a1de18f55c5f2f/route/build/src/main_route.f90#L30) subroutine, then passed to [basin2reach](https://github.com/ESCOMP/mizuRoute/blob/8d487fcfbcbeaf7b5d212a4f29a1de18f55c5f2f/route/build/src/main_route.f90#L143).  It turns out that this change was not needed, but maybe keep this for now.
3. The way to limit outflow from a reach based on the available water storage in the reach is changed in the routing scheme, so that the large negative volume does not happen (see each routing scheme F90 files). Changing default `runoffMin` from 1.e-15 m3/s to zero, caused bigger negative channel volume when it has very small volume and outflow

Because of 3, it slightly change the answer.

### DIFF
--- a/route/build/src/mc_route.f90
+++ b/route/build/src/mc_route.f90
@@ -267,29 +267,29 @@ CONTAINS
  Q(0,1) = rstate%molecule%Q(2) ! outflow at previous time step (t-1)
  Q(1,1) = realMissing ! to be solved
 
- associate(S         => rch_param%R_SLOPE,    & ! channel slope
-           n         => rch_param%R_MAN_N,    & ! manning n
-           bt        => rch_param%R_WIDTH,    & ! channel bottom width
-           bankDepth => rch_param%R_DEPTH,    & ! bankfull depth
-           zc        => rch_param%SIDE_SLOPE, & ! channel side slope
-           zf        => rch_param%FLDP_SLOPE, & ! floodplain slope
-           bankVol   => rch_param%R_STORAGE,  & ! bankful volume
-           L         => rch_param%RLENGTH)      ! channel length
+ associate(r_slope   => rch_param%R_SLOPE,    & ! channel slope
+           r_man_n   => rch_param%R_MAN_N,    & ! manning n
+           r_width   => rch_param%R_WIDTH,    & ! channel bottom width
+           r_depth   => rch_param%R_DEPTH,    & ! bankfull depth
+           side_slope=> rch_param%SIDE_SLOPE, & ! channel side slope
+           fldp_slope=> rch_param%FLDP_SLOPE, & ! floodplain slope
+           r_storage => rch_param%R_STORAGE,  & ! bankful volume
+           rlength   => rch_param%RLENGTH)      ! channel length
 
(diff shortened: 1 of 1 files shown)


route/build/src/dfw_route.f90
```diff
@@ -177,7 +177,7 @@ SUBROUTINE dfw_rch(this,           & ! dfw_route_rch object to bound this proced
    write(iulog,'(A,1X,G15.4)') ' RCHFLX_out(segIndex)%REACH_Q=', RCHFLX_out(segIndex)%ROUTE(idxDW)%REACH_Q
  endif
 
- if (RCHFLX_out(segIndex)%ROUTE(idxDW)%REACH_VOL(1) < 0) then
+ if (RCHFLX_out(segIndex)%ROUTE(idxDW)%REACH_VOL(1) < -1.0e-50_dp) then
```
ekluzek: make a parameter for this like in the last example.

route/build/src/mc_route.f90
```diff
@@ -277,26 +278,26 @@ SUBROUTINE muskingum_cunge(rch_param,     & ! input: river parameter data struct
 
  if (.not. isHW .or. hw_drain_point==top_reach) then
 
-   if (rch_param%RLENGTH > min_length_route) then
+   if (L > min_length_route) then
```
ekluzek: Maybe the associate statement should be RLENGTH so it's easier to figure out it refers to the rch_param object.

route/build/src/dfw_route.f90
nmizukami: replaced with negVolTol(=-1.0e-50_dp), which is defined in public_var.f90

route/build/src/mc_route.f90
nmizukami: done!

PR #609, opened 2026-02-27
refactoring tracer name handling in ctsm/cesm coupling

### LINKED ISSUE #153: Update mizuRoute to handle frozen stream
Update mizuRoute to handle the frozen stream for coupling to CESM.

### LINKED ISSUE #608: Still ice runoff from snowcap is missing
`Flrl_rofi` is imported and stored in `ctl%qsur(:,nice)`, and this is supposed to be put into `ctl%direct(:,nice)` and transferred to the outlet through direct_to_outlet operation. This operation has not been implemented so ice runoff is still missing.

nmizukami:
Now, tracer name including liquid and ice are hard coded and not allow to incorporate other constituents (isotopes, DOC etc). This PR code mode to accommodate additional tracer names in the future.

Also, Now ice flux (from CLM snowcap) is passed to the coupler via direct_to_ocean. 

Fix #608 
Fixes #153

### DIFF
--- a/route/build/cpl/RunoffMod.F90
+++ b/route/build/cpl/RunoffMod.F90
@@ -3,62 +3,67 @@ MODULE RunoffMod
 ! DESCRIPTION:
 ! Module containing data structures for coupler runoff data
 
-  USE nrtype,     ONLY : r8 => dp
+  USE nrtype,     ONLY : r8 => dp, cs => strLen
   USE public_var, ONLY : iulog
   USE public_var, ONLY : integerMissing
-  USE rtmVar,     ONLY : nt_rof
   USE shr_sys_mod,ONLY : shr_sys_abort, shr_sys_flush
 
   implicit none
 
   private
 
-  type, public :: runoff_flow
+  type, public :: rof_control
+    ! tracers
+    integer           :: ntracers = integerMissing   ! number of tracers
+    character(len=cs), allocatable :: tracer_names(:) ! tracer names
+    integer           :: nt_liq                      ! index of liquid tracer in tracer_names
+    integer           :: nt_ice                      ! index of ice tracer in tracer_names
+    logical           :: rof_from_glc                ! if true, will receive liq and ice runoff from glc
+    ! domain and decomposition info
     integer           :: begr,endr        ! local start/stop indices
     integer           :: lnumr            ! local number of catchments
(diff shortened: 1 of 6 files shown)


route/build/cpl/RunoffMod.F90
```diff
@@ -3,82 +3,115 @@ MODULE RunoffMod
...
+    this%qgwl(:,:)       = 0._r8
+    this%discharge(:,:)  = 0._r8
+    this%direct(:,:)     = 0._r8
+    this%volr(:)         = 0._r8
+    this%flood(:)        = 0._r8
+
+  END SUBROUTINE init
+
+  SUBROUTINE init_tracer_names(this, tracer_names)
+    implicit none
```
ekluzek, line 100: Would be good to have a description of what it does.

route/build/cpl/RunoffMod.F90
```diff
@@ -3,82 +3,115 @@ MODULE RunoffMod
...
-
+    ! export variables
     real(r8), pointer :: direct(:,:)      ! coupler return direct flow to ocean [mm/s]
-
     real(r8), pointer :: discharge(:,:)   ! coupler exporting river discharge [m3/s]
     real(r8), pointer :: volr(:)          ! coupler exporting river storage per unit HRU area (m)
     real(r8), pointer :: flood(:)         ! coupler exporting flood water sent back to clm [m3/s]
-  end type runoff_flow
-
-  type(runoff_flow), public :: ctl
```
ekluzek, line 39: This moves it closer to a object oriented way, rather than procedural. So good change.

ekluzek (approved):
@nmizukami and I looked at this together. I had one suggestion to add a description of the new method added. But, even that function has a good name so it's pretty clear what's going on.

This is also moving towards being more object oriented which is a good change. This is more similar to how MOSART is handling ice streams, which is OK for now. More work on tracers will need further changes for generalization.

But, having direct_to_outlet and ice_runoff working are things we need for being coupled to the ocean model in CESM. So a very important update.

route/build/cpl/RunoffMod.F90
nmizukami: Added the description for `init` subroutine and `init_tracer_names` subroutine

PR #611, opened 2026-03-03
Fix locations of drof, srof and xrof

billsacks:
This allows compsets with drof, srof or xrof to be run from standalone mizuRoute checkouts.

The location of xrof is tied to https://github.com/ESCOMP/CMEPS/pull/637 and https://github.com/ESMCI/cime/pull/4946 (but this PR can be merged before or after those because it's unusual to run X compset tests from a standalone mizuRoute checkout).


ekluzek (approved):
Yep, this is fine. We aren't actually using a standalone checkout of mizuRoute for CESM cases right now, so it probably doesn't matter. But, eventually it would be good to get that working, so we might as well bring this in. The same file for CTSM does something similar where it points to the versions for dlnd, slnd, and xlnd.

PR #612, opened 2026-03-06
Need at least one white space before an ampersand

### LINKED ISSUE #610: Continuation line change needed for ifx with intel/2025.3.2
There isn't whitespace before a continuation line in the following file:

``` fortran
src/gamma_func.f90: gammln=tmp+log(stp*(1.000000000190015_dp+&
```

I see the same pattern that Jim identified in CTSM in one file in mizuRoute. So we should fix it in in mizuRoute as well. I haven't verified it fails to compile, but it's most likely a problem.

intel/2025.3.2 seems to have trouble if there isn't whitespace before the continuation line. 
This is similar to this issue in CTSM:

https://github.com/ESCOMP/CTSM/issues/3791

nmizukami:
Add at least one whitespace(s) before a ampersand (&) used for line continuation.

Resolved #610

### DIFF
--- a/route/build/cpl/nuopc/rof_import_export.F90
+++ b/route/build/cpl/nuopc/rof_import_export.F90
@@ -160,7 +160,7 @@ contains
          numflds=fldsFrRof_num, &
          flds_scalar_name=flds_scalar_name, &
          flds_scalar_num=flds_scalar_num, &
-         tag=subname//':mizuRouteExport',&
+         tag=subname//':mizuRouteExport', &
          mesh=Emesh, rc=rc)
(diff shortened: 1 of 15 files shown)


ekluzek (approved):
The changes look cleaner and easier to read, so this is great.

In my testing with CTSM and FATES, I found the problem is more restricted, but it's also not clear what the requirements are either. The main problem in CTSM was that the continuation line was in the middle of a string. And doing that is a bad idea anyway.

But, this is a great thing to do for formatting consistency and readability so great to bring in.

PR #617, opened 2026-03-11
update default control variables for cesm-coupled mode

### LINKED ISSUE #616: Change some of the defaults for the coupled model
Make the defaults as close to the MOSART defaults as we can at this point.

- Change bypass_routing to "direct_to_outlet"
- qgwl_runoff_option to "negative"
- Change for ICE runoff is on
- Turn area correction on?
- Any other defaults that should be turned on for the coupled cases?

nmizukami:
Update default setup for cesm/ctsm coupled mode to the current default MOSART setup.

The following changes:

- `bypass_routing` to "direct_to_outlet"
- `qgwl_runoff_option` to "negative" (this is active if `bypass_routing` is switched to "direct_in_place")
- `ice_runoff` to True
- `dt_qsim` to 3600 

`correct_area` is False now (it is True for MOSART) because this works properly for 0.5 degree grid for mizuRoute.

resolved #616

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -8,8 +8,7 @@ MODULE RtmMod
                           shr_pio_getrearranger, shr_pio_getioroot, shr_pio_getiosys
   USE shr_kind_mod, ONLY: r8 => shr_kind_r8, CL => SHR_KIND_CL
   USE shr_sys_mod,  ONLY: shr_sys_flush, shr_sys_abort
-  USE RtmVar,       ONLY: ice_runoff, &
-                          river_depth_minimum, &
+  USE RtmVar,       ONLY: river_depth_minimum, &
                           nsrContinue, nsrBranch, nsrStartup, nsrest, &
                           cfile_name, coupling_period, nsub, &
                           caseid, brnch_retain_casename, inst_name, &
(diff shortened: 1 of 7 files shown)


nmizukami:
Need to add control variable - ice_runoff so a user can switch on/off for ice runoff (whether track ice separately from liquid or ice is combined into liquid discharge at the outlet

ekluzek (approved):
LGTM

nmizukami:
Hi @ekluzek,  I got restart writing at strange time stamps for testing this. I think it is due to changing time step to 3600, and something not properly set. I will look into this. 

you can see this directory: /glade/derecho/scratch/mizukami/ctsm-mizuRoute/test_f09_f09/run.

Also, I think we will need to make some hybrid parallel work for mizuRoute. I did noticed mainstem domain gets slower (because of more reaches) compared to tributary routing.

You can look inside this: /glade/u/home/mizukami/proj/ctsm-mizuRoute/run_case/test_f09_f09_rHDMAlk/timing/cesm_timing.test_f09_f09_rHDMAlk.5352101.desched1.260312-090040. and compare `route/tributary-route` vs `route/mainstem_route`.  HDMA global should be faster than this (based on the stand-alone run).

PR #618, opened 2026-03-19
Fix for nag compiler

### LINKED ISSUE #615: Build issues with izumi_nag for v3.0.1
in v3.0.1 I get these build errors for SMS_D_Ld5.5x5_amazon_rHDMA.I2000Clm60SpMizGs.izumi_nag.mizuroute-default:

```
Extension: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 98: Line longer than 132 characters
Warning: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 200: Unused dummy variable THIS
Extension: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 296: Line longer than 132 characters
Extension: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 300: Line longer than 132 characters
Extension: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 326: Line longer than 132 characters
Warning: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 403: Unused dummy variable T0
Warning: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 403: Unused dummy variable T1
Warning: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 405: INTEGERMISSING explicitly imported into MC_ROUTE_MODULE but not used
Time(Parsing): User 0.07, System 0.63
Error: /fs/cgd/data0/erik/ctsm_worktree/mizuroute/components/mizuRoute/route/build/src/mc_route.f90, line 192: No spacing specified for X edit descriptor
Time(Semantic Analysis): User 0.00, System 0.00
[NAG Fortran Compiler error termination, 1 error, 28 warnings]
gmake: *** [/fs/cgd/data0/erik/ctsm_worktree/mizuroute/cime/scripts/SMS_D_Ld5.5x5_amazon_rHDMA.I2000Clm60SpMizGs.izumi_nag.mizuroute-default.20260308_165904_ovtoiy/Tools/Makefile:973: mc_route.o] Error 2
```

nmizukami:
1. Need to 1 before X edit descriptor for write format.

Resolve #615

### DIFF
--- a/route/build/cpl/RtmFileUtils.F90
+++ b/route/build/cpl/RtmFileUtils.F90
@@ -95,8 +95,6 @@ contains
      integer,          intent(in)  :: iflag   !0=>abort if file not found 1=>do not abort
 
      ! !LOCAL VARIABLES:
-     integer i               !loop index
-     integer klen            !length of fulpath character string
      logical lexist          !true if local file exists
      !--------------------------------------------------
 
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -7,7 +7,7 @@ MODULE RtmMod
   USE shr_pio_mod,  ONLY: shr_pio_getiotype, shr_pio_getioformat, &
                           shr_pio_getrearranger, shr_pio_getioroot, shr_pio_getiosys
   USE shr_kind_mod, ONLY: r8 => shr_kind_r8, CL => SHR_KIND_CL
-  USE shr_sys_mod,  ONLY: shr_sys_flush, shr_sys_abort
+  USE shr_sys_mod,  ONLY: shr_sys_abort
   USE RtmVar,       ONLY: river_depth_minimum, &
                           nsrContinue, nsrBranch, nsrStartup, nsrest, &
                           coupling_period, nsub, &
@@ -126,25 +126,23 @@ CONTAINS
     !-------------------------------------------------------
     ! mizuRoute time initialize based on time from coupler
     call init_time(ierr, cmessage)
-    if(ierr/=0) then; cmessage=trim(subname)//trim(cmessage); call shr_sys_flush(iulog); call shr_sys_abort(trim(cmessage)); endif
+    if(ierr/=0)then; call shr_sys_abort(trim(subname)//trim(cmessage)); endif
 
     if (masterproc) then
       write(iulog,*) 'define run:'
       write(iulog,*) '   run type              = ',runtyp(nsrest+1)
       write(iulog,*) '   coupling_frequency    = ',coupling_period, '[day]'
       write(iulog,*) '   mizuRoute timestep    = ',dt, '[sec]'
-      call shr_sys_flush(iulog)
+      flush(iulog)
     endif
 
     if (coupling_period <= 0) then
        write(iulog,*) subname,' ERROR mizuRoute coupling_period invalid',coupling_period
-       call shr_sys_flush(iulog)
        call shr_sys_abort( subname//' ERROR: coupling_period invalid' )
     endif
 
     if (dt <= 0) then
       write(iulog,*) subname,' ERROR mizuRoute dt invalid',dt
-      call shr_sys_flush(iulog)
       call shr_sys_abort( subname//' ERROR: mizuRoute dt invalid' )
     endif
 
@@ -154,7 +152,7 @@ CONTAINS
     select case(shr_pio_getioformat(inst_name))
       case(PIO_64BIT_OFFSET); pio_netcdf_format = '64bit_offset'
       case(PIO_64BIT_DATA);   pio_netcdf_format = '64bit_data'
-      case default; call shr_sys_flush(iulog); call shr_sys_abort(trim(subname)//'unexpected netcdf format index')
+      case default; call shr_sys_abort(trim(subname)//'unexpected netcdf format index')
     end select
 
     select case(shr_pio_getiotype(inst_name))
@@ -162,7 +160,7 @@ CONTAINS
       case(pio_iotype_pnetcdf);  pio_typename = 'pnetcdf'
       case(pio_iotype_netcdf4c); pio_typename = 'netcdf4c'
       case(pio_iotype_NETCDF4p); pio_typename = 'netcdf4p'
-      case default; call shr_sys_flush(iulog); call shr_sys_abort(trim(subname)//'unexpected netcdf io type index')
+      case default; call shr_sys_abort(trim(subname)//'unexpected netcdf io type index')
     end select
 
     !pio_numiotasks    = shr_pio_(inst_name)    ! there is no function to extract pio_numiotasks in cime/src/drivers/nuops/nems/util/shr_pio_mod.F90
@@ -183,7 +181,7 @@ CONTAINS
     !-------------------------------------------------------
 
     call init_ntopo_data(iam, npes, mpicom_rof, ierr, cmessage)
-    if(ierr/=0)then; call shr_sys_flush(iulog); call shr_sys_abort(trim(subname)//trim(cmessage)); endif
+    if(ierr/=0)then; call shr_sys_abort(trim(subname)//trim(cmessage)); endif
 
     !-------------------------------------------------------
     ! 4. Define Decomposed domain
@@ -232,11 +230,9 @@ CONTAINS
     end if
 
     if ( any(ctl%gindex(ctl%begr:ctl%endr) < 1) )then
-      call shr_sys_flush(iulog)
       call shr_sys_abort(trim(subname)//"bad gindex < 1")
     endif
     if ( any(ctl%gindex(ctl%begr:ctl%endr) > ctl%numr) )then
-      call shr_sys_flush(iulog)
       call shr_sys_abort(trim(subname)//"bad gindex > max")
     endif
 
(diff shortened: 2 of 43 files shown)


nmizukami:
Hi Erik (@ekluzek).  I was able to build ctsm after the write format fix. But I will clean up including line lengths etc.

nmizukami:
Looks like the issue that cause NAG compiler was [2e3511a](https://github.com/ESCOMP/mizuRoute/pull/618/commits/2e3511a57979d638b11460baa4dcd7258d9341ef). The rest would be warning

Fixed many of F90s but these still include loner lines:

1. ./src/data_assimilation.f90
2. ./src/histVars_data.f90
3. ./src/lake_route.f90
4. ./src/popMetadat.f90 (difficult to fix)
5. ./src/read_streamSeg.f90
6. ./src/write_restart_pio.f90

route/build/cpl/nuopc/rof_comp_nuopc.F90
```diff
@@ -67,7 +66,7 @@ module rof_comp_nuopc
   integer                 :: flds_scalar_num = 0
   integer                 :: flds_scalar_index_nx = 0
   integer                 :: flds_scalar_index_ny = 0
-  integer                 :: flds_scalar_index_nextsw_cday = 0._r8
+  integer                 :: flds_scalar_index_nextsw_cday = 0
```
ekluzek, line 69: Always good to make sure constants to set double precision floats are properly double precision with the _r8 kind at the end.

route/build/cpl/RtmFileUtils.F90
```diff
@@ -95,8 +95,6 @@ subroutine getfil (fulpath, dir, locfn, iflag)
      integer,          intent(in)  :: iflag   !0=>abort if file not found 1=>do not abort
 
      ! !LOCAL VARIABLES:
-     integer i               !loop index
```
ekluzek, line 98: Looks like some of this is general cleanup as well, including removing unused variables. Which is great to do. But, you should mention it in the PR.

route/build/cpl/RtmMod.F90
```diff
@@ -128,7 +126,7 @@ SUBROUTINE route_ini()
     !-------------------------------------------------------
     ! mizuRoute time initialize based on time from coupler
     call init_time(ierr, cmessage)
-    if(ierr/=0) then; cmessage = trim(subname)//trim(cmessage); call shr_sys_flush(iulog); call shr_sys_abort( subname//cmessage ); endif
+    if(ierr/=0) then; cmessage=trim(subname)//trim(cmessage); call shr_sys_flush(iulog); call shr_sys_abort(trim(cmessage)); endif
```
ekluzek: The trim is good practice to always have.

The other change was to remove whitespace around the "=" sign -- was that required for nag?

route/build/src/allocation.f90
```diff
@@ -71,18 +70,18 @@ SUBROUTINE alloc_struct(nHRU,          & ! input: number of HRUs
 
  ! allocate the spatial dimension in all data structures
  allocate(structHRU(nHRU), structHRU2seg(nHRU), structSeg(nSeg), structNTOPO(nSeg), structPFAF(nSeg), stat=ierr)
- if(ierr/=0)then; ierr=20; message=trim(message)//'problem allocating [structHRU,structHRU2seg,structNTOPO,structPFAF]'; return; endif
+ if(ierr/=0)then; message=trim(message)//'problem allocating structHRU, structHRU2seg, structNTOPO or structPFAF'; return; endif
```
ekluzek, line 73: This is done in this and a few below. The error code was getting set to "20" rather than whatever it was beforehand. I'm not sure why that was being done -- but that is a potential change in behavior -- so I assume this was intentional?

route/build/src/network_topo.f90
```diff
@@ -844,7 +835,7 @@ SUBROUTINE reach_mask(desireId,      &  ! input: reach index
  ! check if we actually want the mask
  if(desireId<0)then
 
-  ! ---------- case 1: no mask desired ---------------------------------------------------------------------------------------------------
+  ! ---------- case 1: no mask desired ---------------------------------------------------------------------------------------------
```
ekluzek, line 838: This is really good to make these lines 132 characters long, so that you have a ruler for the rest of the code. I like that.

route/build/src/nr_utils.f90
```diff
@@ -29,12 +29,12 @@ MODULE nr_utils
 END INTERFACE
 
 INTERFACE char2int
-  module procedure :: char2int_1d
-  module procedure :: char2int_2d
+  module procedure char2int_1d
```
ekluzek, line 32: Ahh, it looks like Nag might require this change as well (to remove the ::)?

route/build/src/process_ntopo.f90
```diff
@@ -106,12 +98,17 @@ SUBROUTINE augment_ntopo(nHRU,             & ! input: number of HRUs
  integer(i4b)                                      :: iSeg                 ! indices for stream segment
  real(dp)     , allocatable                        :: seg_length(:)        ! temporal array for segment length
  type(dlength), allocatable                        :: temp_dat(:)          ! temporal storage for dlength data structure
- integer*8                                         :: time0,time1,cr       ! for timing
 
  ierr=0; message='augment_ntopo/'
 
- call system_clock(count_rate=cr)
- call system_clock(time0)
+ ! To time each process, these are what to do
```
ekluzek: First, is uncommenting these lines something that you'll sometimes do in practice? In general I don't like commented out code, but this might be an example of an exception when it's something you do, or show others to do.

Second, I really recommend making use of the GPTL timers for general timing of code. I guess the downside is that it might make the standalone build more complex. But, one way to manage that would be to have a module with subroutines that can do either, but are called in the code. And then an initialization step that can turn them on or off. We could talk more about that...

route/build/cpl/RtmMod.F90
ekluzek: Make it shorter. Makes sense.

You could make a general small error handler something like...

``` fortran
call check_ierr(ierr, cmessage)
```

The the check_ierr subroutine does all the things.

The shr_sys_flush isn't really needed -- because the shr_sys_abort does it as well.

route/build/cpl/nuopc/rof_comp_nuopc.F90
ekluzek: Oh, actually removing it because it's an integer and not a float. Yes, good to do.

route/build/src/allocation.f90
ekluzek: Yes, this was just copied from something else. But, there is no reason to change the ierr value.

route/build/src/nr_utils.f90
ekluzek: It was a warning.

route/build/src/process_ntopo.f90
ekluzek: So do something like:

``` fortran
module mizuRouteTimers

subroutine TimerInit( turn on or off)

subroutine start_timer(description)

subroutine stop_timer(description)
```

Implementation has it figure out if it's running CESM -- it uses GPTL, and if not use the clock.

ekluzek: It would CPP tokens something like...

``` fortran
subroutine start_timer(descriptor)

#ifdef CESMCOUPLED
  call t_startf(description)
#else  
 call system_clock(time0)
 write(*,'(a,1x,1PG15.7,A)') description, ': elapsed time = ', real(time1-time0,kind(dp))/real(cr), ' s'
#endif
```

ekluzek: Or just using GPTL for standalone and CESM if it's already part of the build would be the best.

ekluzek (approved):
Lots's of great stuff here. It'll be nice to work on Izumi with nag. That gives me more confidence in the code. Izumi_nag has often caught legit problems for me with issues in CTSM.

Also good to do the general cleanup of removing unused variables and that sort of thing.

route/build/src/allocation.f90
nmizukami: Yes, ierr is already non-zero integer, which indicates the error. No need to be changed to 20.

route/build/src/nr_utils.f90
nmizukami: I think the correct syntax for generic interface including module procedures is (no `::` needed)

``` fortran
interface generic_name 
  module procedure function_1
  module procedure function_2
  ...
end interface
```

route/build/cpl/RtmMod.F90
nmizukami: removed `shr_sys_flush`, and `trim(subname)//trim(cmessage)` is inserted into `shr_sys_abort` argument

route/build/src/process_ntopo.f90
nmizukami: removed commented code.  if timing is implemented, will use GPTL

route/build/cpl/RtmMod.F90
nmizukami: This has been cleaned up. removed `shr_sys_flush`.  Looks like just `flush` (standard 2003 or later) should be  used. and `shs_sys_abort` does include `flush`.

PR #619, opened 2026-03-28
Flood volume correction

nmizukami:
flood volume was initialized correctly when doing restart when lake is turned on. This happen even when floodplain is off. This does not affect discharge and storage in channel.  

Also, instead of looping through array elements, use where statement to detect negative flow and move them to `direct` array. This seems to be more efficient. 

Turned off floodplain for default as the flood in MOSART is also off.  Made the sign of the flood exported to the coupler negative (as MOSART does), but unclear why


ekluzek (approved):
This is bringing in some important things, like tweaking the tolerance, making sure the sign is right for flooding, and making flooding off the default for coupled cases. I don't see anything to talk about with a quick skim of the changes.

PR #623, opened 2026-04-15
change how restart file is specified

### LINKED ISSUE #621: Add ability to coldstart or use a different restart file for the RUN_TYPE=hybrid option for cesm-coupling
The control file determines the starutp file for mizuRoute to start from (fname_state_in). In the hybrid and branch case (setting of the XML variable RUN_TYPE), though it ignores this variable and uses what's set in the REF XML variables.

There are times when you want to override this behavior, and you want to have the other models do a restart from existing restarts, but you want to provide a different one for mizuRoute and/or do a coldstart for mizuRoute.

I think the way to do this would be to either:

1.) Add an option(s) to the control file

2.) Add an XML option(s) for mizuRoute for cesm-coupling cases

Since, in standalone cases the fname_state_in is always used -- this really only applies to the cesm-coupled case. So it seems like the second one makes the most sense. So an XML variable for mizuRoiute could be added with a name like "USE_CONTROL_FILE_FOR_STARTUP_FILE" or "OVERRIDE_REF_CASE_STARTUP_FILE".

Definition of done:
- [ ] Write a warning to the user when the user has set fname_state_in and it's being overridden and to say they can use the XML variable to change this behavior
- [x] Decide if this should be done for both hybrid AND branch (since branch is usually used to have an exact restart I think the answer is NO for branch)
- [x] Figure out how this should be done, and define what the variable names should be (no names, just use from control file if wet)
- [X] Add an XML option to cime_config/config_component.xml (won't do this)
- [ ] Add a testcase for this in the testmods and testlist
- [ ] Add the handling of it in cime_config/buildnml
- [ ] Make sure this works in the test list

### LINKED ISSUE #641: Bug in handling of restart files for buildnml for coupled cases
The following line in the code is incorrect and needs to be updated to join the rundir rather than the ancil_dir.

https://github.com/ESCOMP/mizuRoute/blob/main/cime_config/buildnml#L245-L246

this was my fault from 3 years ago in 566bcfbab1d90a9554900c5bce62daa55b3d18b2

The code is:

    if fname_state_in is not "empty":
         nmlgen.set_value( "fname_state_in", value=os.path.join( ancil_dir, fname_ntopOld ) )

It must have been copied from a line for fname_ntopOld, and wasn't updated. So it's just been wrong from the beginning.

nmizukami:
Here is how restart file setup works after this PR.

1. get `fname_state_in`
2. if `fname_state_in` in control file is 'STATE_IN_NC' **AND** cesm/ctsm run type is `branch` or `hybrid`, construct restart file based on `RUN_REFCASE` and `RUN_REFDATE`, `RUN_REFTOD` and set the constructed name to `fname_state_in`.
3. if `fname_state_in` in control file is 'STATE_IN_NC' only, set "empty" to `fname_state_in`. (so cold start)
4. neither 2 nor 3 == something other than "empty",  trust user put a valid restart file name in control file. if not, mizuRoute will complains. 

resolve #621
resolve #641

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -236,7 +236,7 @@ def _create_control_files(case, caseroot, srcroot, confdir, inst_string, infile,
                 filename = "%s.mizuroute.r.%s-%s.nc" %(run_refcase, run_refdate, run_tod)
 
             fname_state_in = filename
-         else:
+        else:
             fname_state_in = "empty"
 
     ctl.set( "fname_state_in", fname_state_in )
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -225,20 +225,18 @@ def _create_control_files(case, caseroot, srcroot, confdir, inst_string, infile,
     #----------------------------------------------------
 
     fname_state_in = ctl.get("fname_state_in")
-    if run_type == 'branch' or run_type == 'hybrid':
-        run_refcase = case.get_value("RUN_REFCASE")
-        run_refdate = case.get_value("RUN_REFDATE")
-        run_tod = case.get_value("RUN_REFTOD")
-        filename = "%s.mizuroute%s.r.%s-%s.nc" %(run_refcase, inst_string, run_refdate, run_tod)
-        if not os.path.exists(os.path.join(rundir, filename)):
-            filename = "%s.mizuroute.r.%s-%s.nc" %(run_refcase, run_refdate, run_tod)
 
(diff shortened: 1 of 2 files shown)


ekluzek (approved):
This looks correct to me. Thanks for figuring it out.

nmizukami:
Hi @ekluzek,  lt seems like current PR is not quite working as expected.

PR #632, opened 2026-05-22
change in how inputFileInfo is populated

### LINKED ISSUE #631: long processing time for runoff netCDF input meta.
A stand-alone use (multiple) netCDF(s) for runoff time series forcing.  In pre-processing phase of standalone mode, the metadata (time units, calendar) and time variable in runoff NetCDFs are read in and populate them in data structure - `inputFileInfo`.  This pre-process is done in all the tasks so that all the tasks have the same information. 

However, if a large number of MPI tasks are used (e.g., 512 tasks), this pre-processing takes too much time. I am aware that as a number of tasks increases, the processing time (only runoff meta data processing) increases. 

To fix this, runoff netCDF meta preprocessing needs to be done in a main task and then the information is broadcasted to the other tasks. 

Most of users would need at most 100 tasks for common river network (even high resolution - global MERIT hydro or CONUS NHDPlus 2-3 millions of river/catchments).

nmizukami:
Use main task to read netCDF time variable metadata and time variables and then broadcast them to the other tasks:
This is efficient to populate inputFileInfo especially when a large number of MPI tasks are used, and if >500 tasks are used, still go through quickly compared to many minutes (just go through `inFile_pop` subroutine) for the previous code.

Example:
 
For 32 cores use and importing time metadata for 26yr global 0.5 degree daily runoff (26 annual netCDFs):

previous code: 33.78320 sec (slowest task), 53.40707 sec (slowest task) 
modified code: 9.577683 sec (slowest task), 9.578055 sec (slowest task) 

This does not affect any simulation results or configurations.

resolve #631

### DIFF
--- a/route/build/src/standalone/model_setup.f90
+++ b/route/build/src/standalone/model_setup.f90
@@ -198,7 +198,6 @@ CONTAINS
                                  secprhour, &   ! time conversion factor (hour->sec)
                                  secprday       ! time conversion factor (day->sec)
   USE mpi_utils,           ONLY: shr_mpi_bcast
-  USE mpi_utils,           ONLY: shr_mpi_barrier
 
   ! Argument variables
   character(len=strLen), intent(in)                 :: dir_name         ! the name of the directory that the txt file located
--- a/route/build/src/standalone/model_setup.f90
+++ b/route/build/src/standalone/model_setup.f90
@@ -183,7 +183,8 @@ CONTAINS
                        inputFileInfo,    & ! output: input file information
                        ierr, message)      ! output: error control
 
-  USE globalData,          ONLY: masterproc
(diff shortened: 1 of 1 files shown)


nmizukami:
Hi @ShervanGharari,  

This part of the code was mostly coded by you.  I ran into some issue when I try running using lots of MPI cores. 

Currently the code is reading netcdf meta by all the MPI cores at the same time to populate inputFileInfo, and if using lots of MPI cores, it stuck here (just taking very long time to go through this section by accessing netCDF by many cores). So I changed this routine, so only main core reads the netCDF tasks and then distributes the information to the other cores using MPI function.  This way makes this section faster (actually even using smaller number of cores).   

Changes are not much (just moving around and added indentations so looks lots of changes), but main change is adding mpi_broadcast.

If you can look through this and maybe approve it. Sure, if you have any comments, they are welcome!

PR #634, opened 2026-05-25
Documentation for lakes and reservoir

### LINKED ISSUE #622: Create documentation for the lake model
The lake model is currently NOT documentated in the mizuRoute tech note and needs to be.

ShervanGharari:
lake section updates in technical note. 

Partly resolved #622


ShervanGharari:
@nmizukami, can you point me to the part of the mizuRoute code where the default values for the water management control keys are defined or initialized? This seems like a new development! Thank you.

nmizukami:
> @nmizukami, can you point me to the part of the mizuRoute code where the default values for the water management control keys are defined or initialized? This seems like a new development! Thank you.

flag to do water management or not (this line and a few under this):
https://github.com/ESCOMP/mizuRoute/blob/f500ec3907830851e2eff7b2cfea8a0acfc198f2/route/build/src/public_var.f90#L107

water management files (this line and several under this): 
https://github.com/ESCOMP/mizuRoute/blob/f500ec3907830851e2eff7b2cfea8a0acfc198f2/route/build/src/public_var.f90#L146

These are modified in read_control.f90:
https://github.com/ESCOMP/mizuRoute/blob/f500ec3907830851e2eff7b2cfea8a0acfc198f2/route/build/src/read_control.f90#L139

https://github.com/ESCOMP/mizuRoute/blob/f500ec3907830851e2eff7b2cfea8a0acfc198f2/route/build/src/read_control.f90#L176

nmizukami:
For abstraction/injection, maybe something like "example of river and laker water abstraction is irrigation and domestic or industrial water usage" for tech-note (what this is use for) and maybe how this affects water balance (i.e., limited water is taken out based on existing lake volume).  

for lake: 
https://github.com/ESCOMP/mizuRoute/blob/f500ec3907830851e2eff7b2cfea8a0acfc198f2/route/build/src/lake_route.f90#L177

for river (I think I slightly modified this part, so I can add the description later). the same for the other routing methods
https://github.com/ESCOMP/mizuRoute/blob/f500ec3907830851e2eff7b2cfea8a0acfc198f2/route/build/src/mc_route.f90#L119

I think how to do this is given in user-guide part?.

ShervanGharari:
> mizuRoute/route/build/src/public_var.f90

@nmizukami Thanks — this is very helpful. It looks like some default names were added more recently(?). For example, runoff now has a default name (runoff), or time is now called time, but this is not yet the case for the water management abstraction and target volume variables (which is ok). I'll update the tables accordingly. Thanks again!

[Reference in public_var.f90](https://github.com/ESCOMP/mizuRoute/blob/f500ec3907830851e2eff7b2cfea8a0acfc198f2/route/build/src/public_var.f90?utm_source=chatgpt.com#L126-L154)

ShervanGharari:
@nmizukami Hi Naoki, I think this pull request is ready to be merged. Could you please review and comment? Thank you in advance!

After this PR, I also need to add the Hanasaki formulation to the page, which I will handle in a separate pull request.

nmizukami:
Thanks Shervan (@ShervanGharari).  

Looks good to me in most part. 

One small comment is that groundwater interaction is not explicitly implemented (you mentioned this may be indirectly mimic in water abstraction) as in the water balance equation (equation number is missing btw).  Should we mention that "recharge from or infiltration through lake bed has not been implemented (so now this is ignored).

For the equation number, maybe need to add a label like 

```
.. math::
   :label: lake_water_balance

   \frac{dS}{dt} = I - O + (P - E + G)\,A - F_{a,i}
```

nmizukami:
The edit look good to me!  One more thing -- **Can you add the labels in the equations??** (or I can do later). 

You are right, default variable names in water management netCDF have not been given and maybe can (should) add. this won't affect anyone but if user use the default variables in the input netcdf, don't need to specify them in control file.

ShervanGharari:
> Thanks Shervan (@ShervanGharari).
> 
> Looks good to me in most part.
> 
> One small comment is that groundwater interaction is not explicitly implemented (you mentioned this may be indirectly mimic in water abstraction) as in the water balance equation (equation number is missing btw). Should we mention that "recharge from or infiltration through lake bed has not been implemented (so now this is ignored).
> 
> For the equation number, maybe need to add a label like
> 
> ```
> .. math::
>    :label: lake_water_balance
> 
>    \frac{dS}{dt} = I - O + (P - E + G)\,A - F_{a,i}
> ```

Hi @nmizukami, I am aware of what you mentioned, but I left it this way because the groundwater term (G) can be implicitly incorporated as part of the runoff formulation.

The lake model can operate using only runoff as input, under the assumption that precipitation, evaporation, sublimation over ice-covered lakes, and groundwater exchange are already accounted for in the land surface model and included in the runoff product. In this configuration, the user can switch the lake forcing from separate precipitation/evaporation inputs to runoff-only forcing, which avoids the need to explicitly represent G.

This also avoids increasing model complexity as additional flux terms are introduced. In that sense, G can be considered implicitly represented within mizuRoute through the external forcing design rather than explicitly modeled. 

I already explained this in the lake page, and I can add more info to that:
https://github.com/ESCOMP/mizuRoute/blob/main/route/build/src/read_control.f90#L138

Additionally, I think the control keys for handling precipitation and evaporation should be included in the runoff input configuration:
https://github.com/ESCOMP/mizuRoute/blob/main/route/build/src/read_control.f90#L142-L148
What would you suggest as the best design choice? I think these flags should be added to the runoff input file.

ShervanGharari:
> The edit look good to me! One more thing -- **Can you add the labels in the equations??** (or I can do later).
> 
> You are right, default variable names in water management netCDF have not been given and maybe can (should) add. this won't affect anyone but if user use the default variables in the input netcdf, don't need to specify them in control file.

@nmizukami I have added labels. Is this what you had in mind?

nmizukami:
Yes, the labels work now. All look good to me. I don't have any other comments. I can merge this and you can open another for hanasaki when you get a chance. 

In user-guide section, there is an explanation on handling precipitation and evaporation.

I think I forgot to explain the control key for runoff variable name in input netcdf(s). So I will edit the user-guide input section (section 2.4.2 Runoff file). And I will add evaporation and precipitation for lake (this might be under section 2.7 lake and reservoir).

nmizukami (approved):
Looks good to me!

ShervanGharari:
> Yes, the labels work now. All look good to me. I don't have any other comments. I can merge this and you can open another for hanasaki when you get a chance.
> 
> In user-guide section, there is an explanation on handling precipitation and evaporation.
> 
> I think I forgot to explain the control key for runoff variable name in input netcdf(s). So I will edit the user-guide input section (section 2.4.2 Runoff file). And I will add evaporation and precipitation for lake (this might be under section 2.7 lake and reservoir).

As you prefer, we can either explain runoff, precipitation, evaporation, and the associated scaling factors in Section 2.4, or split the content between Sections 2.4 and 2.7 if that is clearer. Personally, I think it is cleaner to keep all of this in Section 2.4 and then briefly restate the relevant aspects of precipitation and evaporation in Section 2.7 where needed.

ShervanGharari:
> Looks good to me!

Thank you for the review and merge! @nmizukami

Issue #395, opened 2023-06-01
Change format of input control file to a standard format?


ekluzek:
The control file for mizuRoute is in a custom format rather than an industry standard format. When industry standard formats are used (such as XML, JSON, YAML, INI-config, etc.) readers and validators can also be used to handle the file, to read, write, change, and verify syntax. This can be very useful.

In bringing mizuRoute into CESM, we may get pushback unless a standard format is used. I just want to start discussing this to hear how important the file format for the control file is.  If there is flexibility in the mizuRoute community maybe we'll just pick on and convert. If there is a desire to keep the same format we can start preparing our arguments. We could also possibly use different formats for standalone vs cesm-coupled.

We don't need to decide on this anytime soon, but should be thinking about it longer term.

nmizukami:
Personally, I am ok with INI-config or Yaml. These are also more modern? (than namelist - this is ok to me). The format like JSON requires opening and closing brackets and also quote for string, so looks cumbersome and more mistake-prone. 

So like Erik said, this is not urgent, and for now current control file works just fine with cesm coupling.

andywood:
Hi Erik, 
Good topic - it's worth making good choices since the control file is often the most handled in workflows that use a model. 
I think some requirements are to be:  commentable (rules out JSON), easily human readable/editable and concise/non-cluttered (rules out XML), machine-readable/writable w/ common open source libraries & code (eg python), keyword rather than order-controlled, and also ideally handles array and structured/hierarchical entries.  I'm not familiar with the INI format but it looks like it might be suitable and consistency with other CESM models also gets a heavy weight here.  For our new PyGMET program we adopted a TOML format, but if INI looks better I'd be open to switching.  -Andy

ekluzek:
From wikipedia, it looks like INI config files are almost identical to TOML. But, TOML is standardized, so if we convert it would be the leading candidate.

Here's the quote from wikipedia...

"TOML's syntax primarily consists of key = value pairs, [section names], and # (for comments). TOML's syntax somewhat resembles that of .[INI files](https://en.wikipedia.org/wiki/INI_file), but it includes a formal specification, whereas the INI file format suffers from many competing variants."

martynpclark:
Andy's framing of the problem is excellent. I do not have experience with TOML -- on first glace it looks really good.

ShervanGharari:
[...]

PR #646, opened 2026-08-13
feat(config): migrate mizuRoute configuration parsing to native TOML format

johnpaulalex:
### Description of changes
Migrates mizuRoute's configuration file format from legacy space-delimited control files (`.control`) to native TOML (`.toml`). The base CESM 'sample' files are now in TOML format; but CESM user overrides and standalone configs can be specified in either control or TOML format.

This PR is split into six logical commits:

1. `feat(deps): add toml-f Fortran library submodule and build system integration` (2d98395)
   - Adds the `externals/toml-f` submodule and base build system integration for Fortran TOML parsing.

2. `feat(config): migrate mizuRoute configuration parsing to native TOML format` (8989fe6)
   - Migrates default configuration samples (`SAMPLE.toml`, `SAMPLE-coupled.toml`) to TOML format and updates Python `mizuRoute_control.py` using `OrderedDict`.

3. `a3968ac`: `feat(config): add Fortran TOML control file parsing and helper interface`
   - Implements native Fortran TOML control file parsing (`read_control_toml`) and generic file suffix dispatching in `read_control.f90` using `toml-f`.

4. `d9d65ec`: `feat(config): extend Python mizuRoute control parser with conversion CLI and template updates` *(Tweak of commit 8989fe6)*
   - Extends `mizuRoute_control.py` with a bi-directional conversion CLI (`--from_toml`, `--from_control`, `--to_toml`, `--to_control`) and updates CIME `buildnml` template handling.

5. `5f51d97`: `build(cime): integrate toml-f CMake build and MRI archive merging for librof.a` *(Tweak of commit 2d98395)*
   - Updates `cime_config/buildlib` to pre-build `toml-f` via CMake, stage Fortran `.mod` headers, and merge `libtomlf.a` into `librof.a` using GNU `ar -M` MRI scripting.

6. `272c768`: `test(standalone): add automated standalone build and run test script for Derecho (#647)`
   - Adds automated test scripting for standalone build and run verification on NCAR Derecho: route/build/test/test_standalone_derecho.sh.

For CESM, buildnml reads the sample.toml and merges it with user_nl_mizuroute(|_toml), whichever is present, and outputs a .toml file for consumption by fortran.
The fortran library accepts either a toml (CESM or standalone) or legacy file (standalone).

toml<->control scripts: 
* python route/settings/mizuRoute_control.py --from_control mizuRoute_control --to_toml mizuRoute_toml
* python route/settings/mizuRoute_control.py --from_toml mizuRoute_toml --to_control mizuRoute_control
(The script accepts underscores or periods before the file extension)

### Specific notes

**Contributors other than yourself, if any:**
- @johnpaulalex

**Linked issues addressed, if any:**
- Work on #395

**Description of generative AI usage:**
- Google Antigravity was used to write the code and tests, followed by human-guided verification.

### Answer Changes & Scientific Impact
- [x] **Bit-for-Bit (B4B)** with baseline master
- [ ] **Roundoff-level differences only**
- [ ] **Expected Answer Changes (ECA)**

### User Interface & Namelist Changes
- **Namelist / Defaults modified?** Yes (migrated default configuration file to `mizuroute.toml`)
- **XML / Build script changes?** Yes (`cime_config/buildlib`, `cime_config/buildnml`, `.gitmodules`)

### Testing planned or performed, if any:
- [x] Executed pFUnit / CIME regression tests

**CTSM / CESM baseline hash-tag:** `28514dea3fd28a2f820c7ecaa9014f820c1bb220`
**PR branch hash-tag:** `8989fe6`

### Requirements before merge:
- [x] The code in this PR branch builds with no errors.
- [x] The code in this PR branch runs with no errors.
- [x] In-code documentation and Fortran docstrings updated.
- [x] This PR either (a) does not create a need to update documentation or (b) includes required documentation updates. **Which?:** (a) Bug fix / unit test addition; no documentation updates required.


cime_config/buildnml
```diff
@@ -246,24 +246,36 @@ def _create_control_files(case, caseroot, srcroot, confdir, inst_string, infile,
         nmlgen.set_value( "fname_state_in", value=os.path.join( ancil_dir, fname_ntopOld ) )
 
     # Read in the user control file for the case and change settings to it
-    file_src = "user_nl_mizuroute_control"
-    user_ctl_file = os.path.join(caseroot, file_src + inst_string)
-    if ( not os.path.exists( user_ctl_file ) ):
-       safe_copy( os.path.join( srcroot, "cime_config", file_src), user_ctl_file )
+    file_src_toml = "user_nl_mizuroute.toml"
```
ekluzek: Use a underscore rather than a dot to make it match how it's done for the control file.

ekluzek:
```suggestion
    file_src_toml = "user_nl_mizuroute_toml"
```

cime_config/user_nl_mizuroute_toml
ekluzek, line 1: change name as above...

cime_config/user_nl_mizuroute_control
```diff
@@ -2,8 +2,7 @@
 ! This is for changes to the mizuRoute control  file only
 ! Changes to the namelist file need to go in the user_nl_mizuRoute file
 !
```
ekluzek: This should stay the same if the control file is going to be left around.

route/settings/mizuRoute_control.py
ekluzek, line 1: I wonder if this file should be left alone, and a new file mizuRute_toml.py added to handle toml format? Then the control one can be removed once toml format takes over.

cime_config/buildnml
ekluzek: Although there are also advantages to the .toml extension, so it really only matters if others have an opinion on this.

route/build/cpl/RtmVar.F90
```diff
@@ -40,7 +40,7 @@ MODULE RtmVar
   integer,           public            :: rtmhist_ndens  = 1                        ! namelist: output density of netcdf history files
   integer,           public            :: rtmhist_mfilt  = 30                       ! namelist: number of time samples per tape
   integer,           public            :: rtmhist_nhtfrq = 0                        ! namelist: history write freq(0=monthly)
-  character(len=256),public            :: cfile_name     = 'mizuRoute.control'
+  character(len=256),public            :: cfile_name     = 'mizuroute.toml'
```
ekluzek: This can't really be done, until the Fortran can actually read in the TOML format file. So this change should be removed until then.

ekluzek (changes requested):
It looks like this removes the old control files, but also has logic to keep them and just add the new TOML file format. As I understand the transition plan this first step needs the old control files to be kept around until they can be removed when the TOML file reading is brought in.

So I think the way this should go is that at this step it should handle both the control file format and the TOML format. So both are kept around. But, at this step it actually reads in the control file format. But, there can be testing/checking that the python is handling the TOML format correctly.

cime_config/user_nl_mizuroute_control
johnpaulalex: I thought we'd have users specify their overrides in a toml file going forward, and merge that with the case's toml

route/build/cpl/RtmVar.F90
johnpaulalex: meaning https://github.com/ESMCI/cime/pull/5024? How should that work - get the CIME PR and this one in minus this line, and then flip it on in a later PR?

ekluzek:
@johnpaulalex thanks for your work on this. I brought it up with a few people and they agree this is a good thing to work on. One of those is @martynpclark and the library he's using to read TOML file format is the same as the library you are using here. So that's a good sign! This is also something that's too low of a priority for us to work on. But, at the same time it's something that makes a good improvement in usability.

I talked with @nmizukami about this a bit and we do see several things that need to be done. It might be good to have a meeting to go over it.

The most important thing missing is that the standalone build was updated to use the toml-f library. But, the CESM build will also be modified. For that, ccs_config might need changes. And also buildlib under cime_config will need to be updated. The buildlib is currently really simple and it just compiles code in a list of directories. That possibly could be done for toml-f -- but it seems like the toml-f build is really what should be called.

We'd also like to have a path forward that has a transition from control file format to TOML format. So both allowed initially, and then later the control file format removed. A tool to help facilitate that would be good to have is a python tool to convert between the formats. I think that shouldn't be hard to write. But, that would allow users with existing control files to convert the format to TOML and show that the results are identical. They may have a bunch of control files they aleady use, so be able to convert to the new format would help facilitate adoption. And later they start a collection of files in TOML format, that they might need to run with an older version and they'll need it converted backwards. This tool could be built on top of the existing mizuRoute_control.py file in route/settings and hopefully wouldn't need many additions to it.

So let me know what you think about this. I'll put a summary of checkboxes in a comment below to list the things I see.

cime_config/buildnml
ekluzek: Actually thinking about it though. For the user_nl_* files in CESM we never use an extension we only use underscores, so we should probably keep that convention until CESM changes it for the user_nl_* files.

cime_config/user_nl_mizuroute_control
ekluzek: I talk about this below. We thought keeping the ability to read either format would be good for a transition time. But, this was also from my understanding that this PR doesn't change the code to read TOML format -- but a future PR will do that step. So this can't change until the Fortran code is updated to read TOML format anyway.

I do think it's a good plan to do this in a few steps. And it makes sense to me to make the first step be to just add the toml library to the build -- but don't use it. And then the next step is to actually use it.

route/build/cpl/RtmVar.F90
ekluzek: My understanding of this PR is that it brings in the ability to build the toml-f library -- but doesn't change the code to read TOML format (until a fture PR is done). As such the old control file format needs to remain, until the ability to read TOML format is in the Fortran code.

The cime PR just allows cime to do differences on TOML format files. So it's nicely independent of this one. And cime isn't a submodule of mizuRoute, so it doesn't apply here anyway.

ekluzek:
These are the list of things that I see that need to go in here:

- [ ] Add a standalone build/run tester (#647)
- [ ] Make sure the standalone tester works on Derecho with Intel
- [ ] Add the build of the toml-f library to the CESM build
- [ ] Keep the option to use either TOML or control file format
- [ ] Run the "mizu" test list to make sure the CESM tests run as expected
- [ ] Add a tool to convert between control file and TOML format (either direction)

ekluzek:
Workign on a transition plan with @johnpaulalex and @nmizukami 

1. Provide a tool to read and convert between the formats and both can be read in
2. Only TOML format is used to be read in, can still use the tool to covert to TOML, fails if provide old format
3. Eventually remove the tool for conversion

route/settings/SAMPLE-coupled.control
nmizukami, line 1: if legacy control file is kept for a while, this should be also left alone?

route/settings/SAMPLE.control
nmizukami, line 1: if legacy control file is kept for a while, this should be also left alone?

cime_config/user_nl_mizuroute_control
nmizukami, line 1: if legacy control file is kept for a while, this should be also left alone?

cime_config/user_nl_mizuroute_control
nmizukami: Looks like John has added the capability of reading either legacy control file or toml control file, and this PR has done both building toml library and reading the variables in toml in mizuRoute. If so, I am ok with doing with one step. 

But I understand we will keep legacy control file around for a while before removing it.

nmizukami (commented):
I understand that this PR include toml reading capability, so after this is merged, user can start toml version?

but I understand we will keep legacy control file around for a while and a user (especially stand-alone use) has time to transit legacy format to toml. To keep both:
 
For standalone use, we want to keep SAMPLE.control (it is not used in codes, just reference user can look, but maybe just leave it for now)

For cesm coupling use,  user_nl_mizuroute-control, and SAMPLE-coupled.control are left alone to be able to use the legacy one. 

There is one conflict in buildnml to be resolved. 

Also, is this still draft or ready for reviews and merge?  

Overall, looks good to me, and thanks John for this.

cime_config/buildnml
johnpaulalex: I changed it to _toml

cime_config/user_nl_mizuroute_toml
johnpaulalex: done

cime_config/user_nl_mizuroute_control
johnpaulalex: See the PR description for more details but basically:

CESM: sample.toml + user-specific toml|control -> toml file to be sent to fortran
standalone: user-specified toml|control -> either goes to fortran
fortran: accepts toml or control

So users can keep passing in control files for the time being.  Is this right?

route/build/cpl/RtmVar.F90
johnpaulalex: I updated the PR description. The fortran now can read both. lmk if I'm missing something.

route/settings/mizuRoute_control.py
johnpaulalex: If you feel strongly I can do it, but for the converter we'd need something to import both of them - read one, write the other.  And this class is mostly agnostic, it works with the underlying dictionary.  Removing control format now just means removing the read/write control methods in this class.

cime_config/user_nl_mizuroute_control
johnpaulalex: See the PR description. You can still pass in a legacy override file.  I'm not sure it makes sense to keep both user_nl_mizuroute_(control|toml) checked in, since the user should only pass in one.  lmk if you want me to change how the design works.

route/settings/SAMPLE.control
johnpaulalex: See the PR description. I'm not sure it makes sense to keep both SAMPLE.(control|toml) checked in, since the user should only pass in one. but I'm happy to discuss how it could work.

route/settings/SAMPLE-coupled.control
johnpaulalex: See the PR description. I'm not sure it makes sense to keep both SAMPLE-coupled.(control|toml) checked in, since the user should only pass in one. but I'm happy to discuss how it could work.

johnpaulalex:
Hi @nmizukami and @ekluzek , I updated the PR description to describe the various permutations. This should be ready for review. And I resolved the merge conflict.

But it sounds like you might want to keep both a .control and a .toml file checked in, maybe?  Trying to swap this all back into my brain but I think there are 3 places to consider:

1. SAMPLE-coupled_toml. I don't know if that make sense to have control format too since buildnml reads this  by default (and has to choose one file format as the default).
2. user_nl_mizuroute(|_control) - same reasoning as above, there can only be one iin the user's $CASEROOT directory.  That said, I'm realizing this code path is a bit inconsistent in another way: it treats no extension as toml, whereas the other two in this list expect an explicit toml or control extension. Should I make user_nl_mizuroute_toml the default?
3. SAMPLE.(toml|control) - maybe ok to have both formats checked in since iiuc the user has to copy one of these and pass it in expliciitly to the mizuroute binary. Should I keep both of these around?

PR #648, opened 2026-08-31
Pass-through for glacier runoff

### LINKED ISSUE #469: Route new GLC runoff fluxes, or at least ensure that they're zero for now
@mvertens is making changes in CESM so that solid & liquid runoff fluxes from CISM are routed to the runoff model rather than directly to the ocean. The plan in MOSART is for these fluxes to be routed through separately so they can be maintained as separate fluxes sent to the ocean – i.e., MOSART will export separate fluxes-to-ocean for GLC-derived runoff than for LND-derived runoff. Long-term, we want the same behavior in mizuRoute.

If it will be a while – say, post-CESM3 – before this can be put in place, then a short-term solution is to import these two new fluxes (GLC-derived solid & liquid runoff) and add some error-checking to ensure that they're zero, aborting if not.

Note that this is only important for configurations with an evolving ice sheet. However, the tentative plan is for the standard CESM3 configurations to have an evolving ice sheet.

nmizukami:
This PR complete pass-through for glacier runoff:  import from coupler, and go through direct to outlet and export to coupler

Fixes #469

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -502,8 +502,7 @@ CONTAINS
 
         ! Distribute "direct runoff to ocean" to targe reach (i.e., outlet of river network)
         call shr_mpi_sparse_distribute(qSend, commRch(:)%destTask, commRch(:)%destIndex, ctl%direct(:,nt_liq), fillvalue=0._r8)
-
-        call shr_mpi_barrier(mpicom_rof)
+        call shr_mpi_barrier(mpicom_rof) ! AI says no need to call mpi barrier
 
       case default; call shr_sys_abort(trim(subname)//'unexpected bypass_routing_option')
     end select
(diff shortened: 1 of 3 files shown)


route/build/cpl/RtmMod.F90
```diff
@@ -502,8 +502,7 @@ SUBROUTINE route_run(rstwr)
 
         ! Distribute "direct runoff to ocean" to targe reach (i.e., outlet of river network)
         call shr_mpi_sparse_distribute(qSend, commRch(:)%destTask, commRch(:)%destIndex, ctl%direct(:,nt_liq), fillvalue=0._r8)
-
-        call shr_mpi_barrier(mpicom_rof)
+        call shr_mpi_barrier(mpicom_rof) ! AI says no need to call mpi barrier
```
nmizukami, line 505: I believe no need for barrier

route/build/cpl/RtmMod.F90
```diff
@@ -517,8 +516,7 @@ SUBROUTINE route_run(rstwr)
 
     ! Distribute "direct runoff to ocean" to targe reach (i.e., outlet of river network)
     call shr_mpi_sparse_distribute(qSend, commRch(:)%destTask, commRch(:)%destIndex, ctl%direct(:,nt_ice), fillvalue=0._r8)
-
-    call shr_mpi_barrier(mpicom_rof)
+    call shr_mpi_barrier(mpicom_rof) ! AI says no need to call mpi barrier
```
nmizukami, line 519: I believe no need for barrier

nmizukami (commented):
just comments for myself

route/build/cpl/RtmMod.F90
ekluzek: Yes, remove barriers unless there is a really, really good reason for it.

ekluzek (approved):
LGTM. Same as MOSART for the pattern. So should be good.

Removing the barriers might be something you try in a different PR.

PR #650, opened 2026-09-08
unit conversion for direct transfer to outlet

nmizukami:
Runoff imported from coupler (liquid and ice runoff from land-model and glacier runoff from glc model) is given in depth, or kg/m2/s (~=mm/s). Also, routed runoff that is exported to the coupler need to be in kg/m2/s. This quantity is thought as depth per area. So after runoff depth imported into ROF model, depth flux is converted to volume flux using local HRU area to get actual amount of water/ice, then route the volume through the network, or directly transfer to the outlet. Before exporting the routed runoff volume, the volume is converted to depth using the local area at the outlet. 

Currently direct to outlet operation does not do this depth-volume (and volume-depth) conversion and transferred depth directly immediately after the runoff is imported. This causes budget imbalance in ROF model, because the area at upstream HRU and area at outlet location are different. 

This PR fix the error.

### DIFF
--- a/route/build/cpl/RtmMod.F90
+++ b/route/build/cpl/RtmMod.F90
@@ -380,6 +380,7 @@ CONTAINS
     real(r8)                     :: irrig_depth            ! depth of irrigation demand during time step [mm]
     real(r8)                     :: river_depth            ! depth of river water during time step [mm]
     real(r8), allocatable        :: qSend(:)               ! array holding negative lateral flow to be sent to outlet
+    real(r8), allocatable        :: mm_to_m3(:)            ! conversion factor based on local area [m2] from kg/m2(=mm) to m3
     logical                      :: finished               ! dummy arguments (not really used)
     character(len=CL)            :: cmessage               ! error message from subroutines
     integer                      :: ierr                   ! error code
@@ -438,7 +439,9 @@ CONTAINS
     call t_startf('mizuRoute_bypass_route')
 
     allocate(qSend(ctl%lnumr))
+    allocate(mm_to_m3(ctl%lnumr))
     qSend = 0._r8
+    mm_to_m3 = 1.e-3_r8*ctl%area(begr:endr)
 
     select case(trim(bypass_routing_option))
       case('direct_in_place')
@@ -481,6 +484,9 @@ CONTAINS
           ctl%qsur(begr:endr, nt_liq) = 0._r8
         end where
 
+        ! --- convert direct unit (kg/m3/s==mm/s to m3/s)
+        ctl%direct(begr:endr, nt_liq) = ctl%direct(begr:endr, nt_liq)*mm_to_m3(begr:endr)
+
       case('direct_to_outlet')
         ! ----  qgwl [mm/s]
         select case(trim(qgwl_runoff_option))
@@ -502,7 +508,6 @@ CONTAINS
 
         ! Distribute "direct runoff to ocean" to targe reach (i.e., outlet of river network)
         call shr_mpi_sparse_distribute(qSend, commRch(:)%destTask, commRch(:)%destIndex, ctl%direct(:,nt_liq), fillvalue=0._r8)
-        call shr_mpi_barrier(mpicom_rof) ! AI says no need to call mpi barrier
 
       case default; call shr_sys_abort(trim(subname)//'unexpected bypass_routing_option')
     end select
@@ -511,12 +516,12 @@ CONTAINS
 
     call t_startf('mizuRoute_direct_to_outlet_land_ice')
 
-    qSend(:) = 0._r8
-    qSend(begr:endr) = qSend(begr:endr) + ctl%qsur(begr:endr, nt_ice) + ctl%qsub(begr:endr, nt_ice) + ctl%qgwl(begr:endr, nt_ice)
+    qSend(begr:endr) = 0._r8
+    qSend(begr:endr) = ctl%qsur(begr:endr, nt_ice) + ctl%qsub(begr:endr, nt_ice) + ctl%qgwl(begr:endr, nt_ice)
+    qSend(begr:endr) = qSend(begr:endr)*mm_to_m3(begr:endr)
 
     ! Distribute "direct runoff to ocean" to targe reach (i.e., outlet of river network)
     call shr_mpi_sparse_distribute(qSend, commRch(:)%destTask, commRch(:)%destIndex, ctl%direct(:,nt_ice), fillvalue=0._r8)
-    call shr_mpi_barrier(mpicom_rof) ! AI says no need to call mpi barrier
 
     ! Set ctl%qsur, ctl%qsub and ctl%qgwl to zero for nt_ice
     ctl%qsur(:,nt_ice) = 0._r8
@@ -527,9 +532,14 @@ CONTAINS
 
     call t_startf('mizuRoute_direct_to_outlet_glc_runoff')
     if (ctl%rof_from_glc) then
+      qSend(begr:endr) = 0._r8
+      qSend(begr:endr) = ctl%qglc_liq(begr:endr)*mm_to_m3(begr:endr)
       ! Distribute "direct runoff to ocean" to targe reach (i.e., outlet of river network)
-      call shr_mpi_sparse_distribute(ctl%qglc_liq(:), commRch(:)%destTask, commRch(:)%destIndex, ctl%direct_glc(:,nt_liq), fillvalue=0._r8)
-      call shr_mpi_sparse_distribute(ctl%qglc_ice(:), commRch(:)%destTask, commRch(:)%destIndex, ctl%direct_glc(:,nt_ice), fillvalue=0._r8)
+      call shr_mpi_sparse_distribute(qSend, commRch(:)%destTask, commRch(:)%destIndex, ctl%direct_glc(:,nt_liq), fillvalue=0._r8)
+
+      qSend(begr:endr) = 0._r8
+      qSend(begr:endr) = ctl%qglc_ice(begr:endr)*mm_to_m3(begr:endr)
+      call shr_mpi_sparse_distribute(qSend, commRch(:)%destTask, commRch(:)%destIndex, ctl%direct_glc(:,nt_ice), fillvalue=0._r8)
     else
       ctl%direct_glc(:,:) = 0._r8
     end if
@@ -539,6 +549,7 @@ CONTAINS
     call t_startf('mizuRoute_mapping_runoff')
 
     ! Transfer actual irrigation rate [mm/s] to river segment
+    ! unit conversion mm/s to m3/s is done in mizuRoute code
     if (masterproc) then
       if (nRch_mainstem > 0) then ! mainstem
         call basin2reach(ctl%qirrig_actual(1:nHRU_mainstem), NETOPO_main, RPARAM_main, flux_wm_main, &
@@ -658,7 +669,8 @@ CONTAINS
                                    offset)
 
     ! Descriptions: get exporting variables
-    !  - discharge [kg/m2/s]=[mm/s]
+    !  - discharge [kg/m2/s = mm/s]
+    !  - direct runoff transfer [mm/s]
     !  - water volume per HUC area [m]
     !  - flood [mm/s]
     !  NOTE 1) offset (optional input): in main processor, index for ctl variable is based on 1 through nHRU_mainstem+nHRU_trib
@@ -709,7 +721,7 @@ CONTAINS
         end if
         if (update_q) then
           if (NETOPO_in(iRch)%DREACHI==-1 .and. NETOPO_in(iRch)%DREACHK<=0) then ! if reach is the outlet
-            ctl%discharge(ix,ctl%nt_liq) = RCHFLX_in(iRch)%ROUTE(iRoute)%REACH_Q* NETOPO_in(iRch)%HRUWGT(iHru)/ctl%area(ix)/0.001_r8
+            ctl%discharge(ix,ctl%nt_liq) = RCHFLX_in(iRch)%ROUTE(iRoute)%REACH_Q* NETOPO_in(iRch)%HRUWGT(iHru)/(ctl%area(ix)*0.001_r8)
           end if
         end if
         if (update_fld) then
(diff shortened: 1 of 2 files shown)


route/build/cpl/RtmMod.F90
```diff
@@ -511,12 +516,12 @@ SUBROUTINE route_run(rstwr)
 
     call t_startf('mizuRoute_direct_to_outlet_land_ice')
 
-    qSend(:) = 0._r8
-    qSend(begr:endr) = qSend(begr:endr) + ctl%qsur(begr:endr, nt_ice) + ctl%qsub(begr:endr, nt_ice) + ctl%qgwl(begr:endr, nt_ice)
+    qSend(begr:endr) = 0._r8
```
ekluzek: In the line below, you noticed that adding in qSend that had been set to zero, doesn't need to be done. The same here, there isn't a reason to set qSend to zero, when the next line defines the whole array. So might as well remove the initial setting to zero. Setting to zero might be useful to do if the next line skipped some of the points in the array, but it doesn't so might as well remove.

route/build/cpl/RtmMod.F90
```diff
@@ -380,6 +380,7 @@ SUBROUTINE route_run(rstwr)
     real(r8)                     :: irrig_depth            ! depth of irrigation demand during time step [mm]
     real(r8)                     :: river_depth            ! depth of river water during time step [mm]
     real(r8), allocatable        :: qSend(:)               ! array holding negative lateral flow to be sent to outlet
+    real(r8), allocatable        :: mm_to_m3(:)            ! conversion factor based on local area [m2] from kg/m2(=mm) to m3
```
ekluzek: I wonder about changing the name from  mm_to_m3 -- to something to help clarify that it includes the area? At first I was thinking this is just 1.e-3 and didn't need to be an arrary. But, since it has area it does need to be an array. And it's good to do this once, and then use it multiple times.

But, actually might might want to save it as a module variable, and only do it at initialization so it isn't done each time-step.

route/build/cpl/RtmMod.F90
ekluzek:
```suggestion
```

route/build/cpl/RtmMod.F90
```diff
@@ -527,9 +532,14 @@ SUBROUTINE route_run(rstwr)
 
     call t_startf('mizuRoute_direct_to_outlet_glc_runoff')
     if (ctl%rof_from_glc) then
+      qSend(begr:endr) = 0._r8
```
ekluzek:
```suggestion
```

route/build/cpl/RtmMod.F90
```diff
@@ -527,9 +532,14 @@ SUBROUTINE route_run(rstwr)
...
+      qSend(begr:endr) = ctl%qglc_liq(begr:endr)*mm_to_m3(begr:endr)
       ! Distribute "direct runoff to ocean" to targe reach (i.e., outlet of river network)
-      call shr_mpi_sparse_distribute(ctl%qglc_liq(:), commRch(:)%destTask, commRch(:)%destIndex, ctl%direct_glc(:,nt_liq), fillvalue=0._r8)
-      call shr_mpi_sparse_distribute(ctl%qglc_ice(:), commRch(:)%destTask, commRch(:)%destIndex, ctl%direct_glc(:,nt_ice), fillvalue=0._r8)
+      call shr_mpi_sparse_distribute(qSend, commRch(:)%destTask, commRch(:)%destIndex, ctl%direct_glc(:,nt_liq), fillvalue=0._r8)
+
+      qSend(begr:endr) = 0._r8
```
ekluzek:
```suggestion

```

route/build/cpl/RtmMod.F90
```diff
@@ -511,12 +516,12 @@ SUBROUTINE route_run(rstwr)
 
     call t_startf('mizuRoute_direct_to_outlet_land_ice')
 
-    qSend(:) = 0._r8
-    qSend(begr:endr) = qSend(begr:endr) + ctl%qsur(begr:endr, nt_ice) + ctl%qsub(begr:endr, nt_ice) + ctl%qgwl(begr:endr, nt_ice)
+    qSend(begr:endr) = 0._r8
+    qSend(begr:endr) = ctl%qsur(begr:endr, nt_ice) + ctl%qsub(begr:endr, nt_ice) + ctl%qgwl(begr:endr, nt_ice)
+    qSend(begr:endr) = qSend(begr:endr)*mm_to_m3(begr:endr)
```
ekluzek: You could include this in the line above, by adding paranthesis and then doing the multiply around the whole thing. Technically that might make a tiny difference to the optimization of the code -- but it's also something that the compiler might do for you. So it mainly depends on what you think is more readable.

In terms of readability I can see it either way, with a slight preference to do it in one line.

route/build/cpl/RtmMod.F90
```diff
@@ -639,6 +650,11 @@ SUBROUTINE route_run(rstwr)
       call get_river_export_data(NETOPO_trib, RCHFLX_trib)
     end if
 
+    ctl%direct(:,ctl%nt_liq) =ctl%direct(:,ctl%nt_liq) / (ctl%area(:)*0.001_r8)
```
ekluzek: Wait all of these should just use mm_to_m3 right?

ekluzek:
```suggestion
    ctl%direct(:,ctl%nt_liq) =ctl%direct(:,ctl%nt_liq) / mm_to_m3(:)
```

route/build/cpl/RtmMod.F90
```diff
@@ -709,14 +726,15 @@ SUBROUTINE get_river_export_data(NETOPO_in, RCHFLX_in, &
         end if
         if (update_q) then
           if (NETOPO_in(iRch)%DREACHI==-1 .and. NETOPO_in(iRch)%DREACHK<=0) then ! if reach is the outlet
-            ctl%discharge(ix,ctl%nt_liq) = RCHFLX_in(iRch)%ROUTE(iRoute)%REACH_Q* NETOPO_in(iRch)%HRUWGT(iHru)/ctl%area(ix)/0.001_r8
+            ctl%discharge(ix,ctl%nt_liq) = RCHFLX_in(iRch)%ROUTE(iRoute)%REACH_Q* NETOPO_in(iRch)%HRUWGT(iHru)/(ctl%area(ix)*0.001_r8)
```
ekluzek:
```suggestion
            ctl%discharge(ix,ctl%nt_liq) = RCHFLX_in(iRch)%ROUTE(iRoute)%REACH_Q* NETOPO_in(iRch)%HRUWGT(iHru)/mm_to_m3(ix)
```

route/build/cpl/RtmMod.F90
ekluzek: It's really area divided by 1000 (the conversion between mm to m). So it's sort of km*m which is just a strange unit of area. So area_km_m? It looks like there is a convention to use a _conv suffix to denote conversions (or as a prefix, it's done both ways), so maybe discharge_to_volume_conv or conv_discharge_to_volume or convDischarge2Volume? They all seem to follow other examples in the code.

ekluzek: In terms of making this only done at initialization. I suggest moving this to become a private module variable at the top of the module. Then the allocation and calculation would be done in route_ini.

route/build/cpl/RtmMod.F90
```diff
@@ -438,7 +439,9 @@ SUBROUTINE route_run(rstwr)
     call t_startf('mizuRoute_bypass_route')
 
     allocate(qSend(ctl%lnumr))
+    allocate(mm_to_m3(ctl%lnumr))
```
ekluzek: Move this to route_ini

route/build/cpl/RtmMod.F90
```diff
@@ -438,7 +439,9 @@ SUBROUTINE route_run(rstwr)
...
     qSend = 0._r8
+    mm_to_m3 = 1.e-3_r8*ctl%area(begr:endr)
```
ekluzek: Move to route_ini

route/build/cpl/RtmMod.F90
ekluzek: Same for the lines below as well..

route/build/cpl/RtmMod.F90
```diff
@@ -709,14 +726,15 @@ SUBROUTINE get_river_export_data(NETOPO_in, RCHFLX_in, &
...
-            ctl%discharge(ix,ctl%nt_liq) = RCHFLX_in(iRch)%ROUTE(iRoute)%REACH_Q* NETOPO_in(iRch)%HRUWGT(iHru)/ctl%area(ix)/0.001_r8
+            ctl%discharge(ix,ctl%nt_liq) = RCHFLX_in(iRch)%ROUTE(iRoute)%REACH_Q* NETOPO_in(iRch)%HRUWGT(iHru)/(ctl%area(ix)*0.001_r8)
           end if
         end if
         if (update_fld) then
           ctl%flood(ix) = RCHFLX_in(iRch)%ROUTE(iRoute)%FLOOD_VOL(1)*NETOPO_in(iRch)%HRUWGT(iHru)/ctl%area(ix)/coupling_period
         end if
       end do
     end do
+
```
ekluzek, line 734: Note, there isn't a deallocate for either mm_to_m3 nor qSend. By moving mm_to_m3 to a module variable done in route_ini, you don't really need a deallocate for it. But, qSend does NEED a deallocate. As it's a memory leak and creating memory each time route_run is called.

route/build/cpl/nuopc/rof_import_export.F90
```diff
@@ -221,7 +221,7 @@ subroutine realize_fields(gcomp, Emesh, flds_scalar_name, flds_scalar_num, corre
       nullify(dataptr)
     end if
 
-    if (fldchk(importState, 'Fgrg_rofl') .and. fldchk(importState, 'Fgrg_rofl')) then
+    if (fldchk(importState, 'Fgrg_rofl') .and. fldchk(importState, 'Fgrg_rofi')) then
```
ekluzek, line 224: Ahhh, nice catch.

route/build/cpl/RtmMod.F90
ekluzek: I made this into an issue #651

ekluzek (changes requested):
This is great. I'm so glad you figured this out. It sounds like this fixes the remaining issues for the budget running coupled, which is really awesome.

Since, this has been in place since the beginning of coupling, I suggest you make this into an issue that this PR resolves. That way the issue will be easier to find in the future.

I do have several suggestions that I think are worthwhile to do. I think moving the allocation and setting of mm_to_m3 to route_ini is worthwhile to do. And removing some of the initiialization of qSend to zero is good. And there are some instances where mm_to_m3 should be used and isn't. Those are all worthwhile.

I wondered about changing the name of mm_to_m3, and have some suggestions that follow other units conversion variables in the code.

I also noticed #651 in looking this over, which would be good to fix.

Anyway, great work. Let me know if you want to discuss any of this.

nmizukami:
> This is great. I'm so glad you figured this out. It sounds like this fixes the remaining issues for the budget running coupled, which is really awesome.
> 
> Since, this has been in place since the beginning of coupling, I suggest you make this into an issue that this PR resolves. That way the issue will be easier to find in the future.
> 
> I do have several suggestions that I think are worthwhile to do. I think moving the allocation and setting of mm_to_m3 to route_ini is worthwhile to do. And removing some of the initiialization of qSend to zero is good. And there are some instances where mm_to_m3 should be used and isn't. Those are all worthwhile.
> 
> I wondered about changing the name of mm_to_m3, and have some suggestions that follow other units conversion variables in the code.
> 
> I also noticed #651 in looking this over, which would be good to fix.
> 
> Anyway, great work. Let me know if you want to discuss any of this.

Hi Erik (@ekluzek), yes, I can talk about what i found out.   This is definitely "one" of rof issues that affected budget table. and I am rerunning a few years to get the latest run and see how the budget table looks like.  Now this fixes transfer all the ice runoff that is imported from land and glacier to the couplers after direct to outlet, without ROF storing or eliminating. But not necessarily reducing entire CESM wide error. 

For CESM 

LND -> cpl -> rof -> cpl -> ocn. 

So mass transfer _through_ each coupler may lose/gain mass if input data (mesh and mapping) is something wrong and error show up in budget table. 

For this PR, I just fixed a part within rof, and I suspect there is still issue rof-to-ocn coupling part (mapping rof runoff to ocean discharge)

route/build/cpl/RtmMod.F90
nmizukami: is local allocatable array in subroutine automatically deallocated when the subroutine exits, as long as it does not have the SAVE attribute?  Yes, I understand if allocatable array is moved to module level and allocated in subroutine, it remain allocated.

route/build/cpl/RtmMod.F90
nmizukami: I am thinking of using `depth_to_vol` instead of `mm_to_m3`?  

kg/m2 (mass per unit area) is mm if assuming water density is 1,000 kg/m³. 
This factor is meant to convert depth [mm] to volume [m³.] by first converting mm to m (multiplying 0.001) then multiplying area [m2]. That is where 0.001*area(:) is coming from

route/build/cpl/RtmMod.F90
nmizukami: Removed this line

route/build/cpl/RtmMod.F90
nmizukami: Removed this line

route/build/cpl/RtmMod.F90
nmizukami: `depth_to_vol` is used now

route/build/cpl/RtmMod.F90
nmizukami: I changed `mm_to_m3` to `depth_to_vol`, and then move it to module level variable. Also initialize this in `route_ini` after `ctl%area` is populated.

route/build/cpl/RtmMod.F90
nmizukami: Yes, I combined sum of three runoff and unit conversion into one line

route/build/cpl/RtmMod.F90
nmizukami: Use `depth_to_vol` for volume-depth conversion

nmizukami:
There is potentially large bug in negative runoff direct to outlet. I did not covert negative runoff from depth to volume before moving it to the outlet using `shr_mpi_sparse_distribute`.  Commit [8fd8181](https://github.com/ESCOMP/mizuRoute/pull/650/commits/8fd81813bd3161645940cbd6dabff5af38ea5f95) fixed this. 

I will need to re-run this for testing.

route/build/cpl/RtmMod.F90
nmizukami: I think this is removed

route/build/cpl/RtmMod.F90
nmizukami: For now, I will resolve this.

ekluzek (approved):
LGTM!

